PR #5, opened 2020-09-24
Disable right click

Tokikko:
Right click on child image is disabled. This still can not prevent someone from downloading it but it will make things a bit more complex for those who do not know.

### DIFF
--- a/public/clientJS/wishCard.js
+++ b/public/clientJS/wishCard.js
@@ -1,0 +1,19 @@
+// check clicked element and disable right click if its the element that contains child image
+function preventChildImageContextMenu(e) {
+    // class name of the element that contains the child image
+    let childImageElementClassName = "card-img-top"; 
+    
+    // take clicked html element
+    let firstElementChild = e.path[0];
+    
+    // check if it exists
+    if (firstElementChild) {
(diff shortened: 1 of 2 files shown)


stacysealky (changes requested):
This is great and thanks for your contribution. Could you save the <script> part into a seperate js file and link it? Just trying to keep things decluttered. You can put the separate js file inside '/controllers' :)

gnorbsl:
I'm not sure if putting frontend scripts in controllers is a good idea @stacysealky 

I think since this is a frontend change, the client won't even have access to controllers. Usually controller is reserved for the backend they handle all the logic from the API

I would suggest to put it in the public folder inside /js/ maybe

Tokikko:
Yeah i agree with @gnorbsl . Better to put them inside a js file, controllers are usually meant to be named for the backend. Can i create/put into a js like it was suggested?

stacysealky:
@Tokikko @gnorbsl You are right! There's a ClientJS folder inside public directory. Maybe put it there?

Tokikko:
Ok, i will put it there then.

Tokikko:
Done

PR #6, opened 2020-09-24
Added docker compose for mongo and fake test email

gnorbsl:
Mongo can now be started with docker, this way no one has to install mongo on their machine.

I also added fake email transports so if emails have to be tested we won't use real ones until necessary.

Removed a few dotenv requires since it only has to be called once in the app.js and then works everywhere

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,5 +11,5 @@ uploads
 node_modules
 
 # Ignore files related to API keys
-*.env
+
 config/config.env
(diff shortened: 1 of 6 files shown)


gnorbsl:
removed an email test that I forgot

stacysealky:
In app.js you set path: './config/dev-config.env' but you named the file /config/example-config.env. Was this intentional? 
Also, SESS_SECRET is displaying in your new .env. Is that ok for security?

gnorbsl:
Good catch I removed it

I thought about adding a env file for production and dev, but since everyone sahould just work with whatever is inside example-config.env as their config.env I scrapped it.

SESS secret is newly generated so not the same as your live config. Since this is mostly for local tests there is no real risk witjh exposing it.

PR #8, opened 2020-09-25
Edit about me info

Tokikko:
Added popup that allows user to edit about me section. Created new endpoint  route (put  /users/profile), extended document model with the required field, added pencil icon from https://material.io/resources/icons/?style=baseline


gnorbsl:
Have you checked if the text in about me allows html or js? I can see a potential xss attack here. Havent tested it yet though.

Tokikko:
I think it should be text only, but im not that well versed with it. You can check and report back im curious now :)

PR #16, opened 2020-09-28
Verified Agencies

Jacobjeevan:
Added logic to display wish card form only to verified agencies.

- To enable it, users with admin role will have to manually set 'isVerified' property for agencies on mongoDB
- We can add logic to handle this directly (within the site) in the future

To facilitate rendering based on verification property, few other modifications were also necessary:
- Attach agency id to request session (similar to user), as well as added logic to fetch/populate agency information (similar to user information).


stacysealky:
> Added logic to display wish card form only to verified agencies.
> 
> * To enable it, users with admin role will have to manually set 'isVerified' property for agencies on mongoDB
> * We can add logic to handle this directly (within the site) in the future
> 
> To facilitate rendering based on verification property, few other modifications were also necessary:
> 
> * Attach agency id to request session (similar to user), as well as added logic to fetch/populate agency information (similar to user information).

gnorbsl (changes requested):
can you use brackets for if statements? or at least put them in the same line. users.js line 88 for example.

also users.js line 188 wrong intendation I think?

I would also suggest that before we put that in place we add a frontend notification for the agency that they have to be approved, maybe in an welcome email after email verification. otherwise it'll look like stuff is not working

app.js
```diff
@@ -63,8 +63,7 @@ mongoose.connect(process.env.MONGO_URI, options, (err, database) => {
 
 //IMPORT MODELS
 const User = require('./models/User');
-const Agency = require("./models/Agency");
-
+con
```
gnorbsl: This has to be changed too. Not sure what went wrong here

gnorbsl (changes requested):

routes/users.js
```diff
@@ -84,9 +84,10 @@ router.get('/login', redirectProfile, (req, res) => {
 // TODO: add conditions to check userRole and limit 'createWishCard' access to 'partners' only
 router.get('/profile', redirectLogin, async (req, res) => {
 	try {
-		res.render('profile', {
-			user: res.locals.user
-		});
+		let params = { user: res.locals.user };
+		if (res.locals.agency)
```
gnorbsl, line 88: please put them in the same line or use brackets on those if statements

routes/users.js
```diff
@@ -171,17 +172,20 @@ router.post('/agency', async (req, res) => {
...
 		agencyWebsite,
 		agencyPhone,
-		agencyBio
+		agencyBio,
+		accountManager: req.session.userId
 	});
 	try {
 		await newAgency.save();
+		var agencyId = mongoose.Types.ObjectId(newAgency._id);
+    	req.session.agencyId = agencyId;
```
gnorbsl, line 188: Intendation wrong?

gnorbsl (changes requested):
test

Jacobjeevan:
Thank you for the comments :+1: 

Fixed the formatting issues.

Great point about displaying a message for unverified users. Here are the changes I made:

- Added logic to display a message for unverified agencies
- Added logic to hide the create wishcard button for unverified agencies (form was already hidden, forgot to add logic for button)

For profile page:

- Earlier version didn't handle all cases (ex: it only handled agency info for newly registered users)
		- Current version takes care of both new and old users

App.js:

Middleware for agency info extraction is strictly not necessary (as of now), since only the profile page uses the information. But I kept it for future use case. Could remove it if needed.

app.js
```diff
@@ -42,78 +42,85 @@ const port = 8081;
 
 //LOAD CONFIG.ENV VARS
 dotenv.config({
-	path: './config/config.env'
+  path: "./config/example-config.env",
```
gnorbsl, line 45: This should be config.env

routes/users.js
```diff
@@ -1,272 +1,277 @@
...
-			error: err
-		}));
-	}
-});
+    // update user and add aboutMe
+    User.updateOne(
+      { _id: candidate._id },
+      { aboutMe: aboutMe },
+      { multi: true },
+      function (err, numberAffected) {}
```
gnorbsl, line 145: I think we should not ignore a failing update, it looks like even with a failed update here we return a 200? since the err is unused and will not be catched by the catch?

routes/users.js
```diff
@@ -1,272 +1,277 @@
...
+    accountManager: req.session.userId,
+  });
+  try {
+    await newAgency.save();
+    var agencyId = mongoose.Types.ObjectId(newAgency._id);
+    req.session.agencyId = agencyId;
+    console.log("agency data saved");
+    return res.send("/users/profile");
+  } catch (err) {
+    console.log(err);
```
gnorbsl, line 205: Without a response I think as soon as this fails the client gets a timeout? we should at least throw a 500 here

routes/users.js
```diff
@@ -1,272 +1,277 @@
...
+      await newUser.save();
+      //trying to add a second step here
+      //if the userRole is partner then redirect to agency.ejs then profile.ejs
+      if (newUser.userRole == "partner") {
+        return res.send("/users/agency");
+      } else {
+        return res.send("/users/profile");
+      }
+    } catch (err) {
+      console.log(err);
```
gnorbsl, line 243: error will time out again I assume?

gnorbsl (changes requested):
A few things: its a pain in the ass to find changes if you auto fomrat everything with double ticks instead of single ticks :D 

It might make sense to add linting to the project

A few missing responses but looks good otherwise

gnorbsl:
Can you check if the request times out if the update fails at the places I marked?

routes/users.js
Jacobjeevan: Removing it since it doesn't do anything (the catch catches the error); I tried adding doing it within the function (customized message, but the catch returns the error instead).

app.js
Jacobjeevan: Yeah, will fix that. I was switching between config and example. We should handle with environment variables :thinking:

Jacobjeevan: I added a local option for scripts (allows for easier switching between environments). Did that rather than using dev since dev requires local mongoDB instance.

routes/users.js
Jacobjeevan: Yep, added logic to handle it.

Jacobjeevan:
Sorry about the formatting lol. Yeah, adding linting would be nice.

I added logic to handle the missing cases :thinking: (added local option to scripts for switching between environments). Didn't want to modify dev since it requires a local mongoDB instance.

gnorbsl:
> Sorry about the formatting lol. Yeah, adding linting would be nice.
> 
> I added logic to handle the missing cases 🤔 (added local option to scripts for switching between environments). Didn't want to modify dev since it requires a local mongoDB instance.

Are you not working with a local mongo instance? if you install docker you can spin up a local mongo instance pretty quickly

Jacobjeevan:
> > Sorry about the formatting lol. Yeah, adding linting would be nice.
> > I added logic to handle the missing cases thinking (added local option to scripts for switching between environments). Didn't want to modify dev since it requires a local mongoDB instance.
> 
> Are you not working with a local mongo instance? if you install docker you can spin up a local mongo instance pretty quickly

I am; I should have clarified - I meant that I kept it that way for others who might not want to install mongo. :thinking: Oh, yeah we have docker setup now, don't we? Forgot about that. Thanks, will do that :D

gnorbsl (approved):
looks good to me

PR #28, opened 2020-10-01
Toast modifications

Tokikko:
Moved css nad js files required for toas into footer since they are used everywhere on the site.
Wrapped the call to make toast inside a function since the design is the same for every page, only the message is changing.
Replace alerts with toast messages

### DIFF
--- a/public/clientJS/agency.js
+++ b/public/clientJS/agency.js
@@ -18,7 +18,7 @@ $(document).ready(function () {
       },
       statusCode: {
         409: function (responseObject, textStatus, jqXHR) {
-          alert(responseObject.responseText);
+          showToast(responseObject.responseText);
         },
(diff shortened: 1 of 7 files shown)


stacysealky:
made small changes to the css file and footer template. 
I think header should have css links instead of footer, just for the loading order.
lmk if this is ok

also add this 
<link rel="stylesheet" type="text/css" href="https://cdn.jsdelivr.net/npm/toastify-js/src/toastify.min.css">
<link rel="stylesheet" type="text/css" href="../css/toast.css">

to the header.ejs template please.

Thanks and great work!

Tokikko:
Did you make those changes already?

stacysealky:
> Did you make those changes already?

yeah, you ok with it?

Tokikko:
Yeah, feel free to merge :)

PR #29, opened 2020-10-01
Messaging feature

Jacobjeevan:
Apologize for all the formatting differences. I added prettier (for linting) to packages, so shouldn't happen again.

- wishCardFullPage.ejs
	- Dropdown -> select
	- Allows user to select a message and send it
	- Displays existing messages, along with User it's from

- wishCardFullPage.js
	- Call to backend for submitting the message
	- To Do: prettify error messages/alerts

- wishCards.js route
	- Modified wishcard get (by id) method to populate messages and as well user.
	- Slight modifications to post /message for updating the wishcards.

Additional comments:

For a guest, clicking on view more on wish card page leads to login page. Alternatively, we could display an alert instead (or maybe do a combination of both).

For now the population approach works, but it might not scale well (performance wise, especially if we have a lot of cards). But, we can worry about that later :D

- Changed AWS bucket name to refer to config file


routes/wishCards.js
```diff
@@ -10,201 +10,259 @@ const router = express.Router();
...
-        res.status(400).send(JSON.stringify({
-            success: false,
-            error: error
-        }));
-    }
+  try {
+    const results = await WishCard.find({
+      wishItemName: {
+        $regex: req.body.wishitem,
+        $options: 'i',
```
gnorbsl, line 127: double ticks got removed here is that an issue?

views/wishCardFullPage.ejs
```diff
@@ -8,6 +8,10 @@
     <!-- ** NOTE: CDN Bootstrap.js, Popper.js, and jQuery are moved to the bottom so the page loads faster-->
     <%- include('./templates/header')-%>
     <link rel="stylesheet" href="../css/wishCardFull.css">
+    <!-- jquery -->
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
+    <!-- js scripts -->
+    <script src="../clientJS/wishCardFullPage.js"></script>
```
gnorbsl, line 14: we dont need double dots, /clientJS/ should be enough and wont cause issues in nested routes, double dots will.

views/wishCardFullPage.ejs
```diff
@@ -40,44 +44,39 @@
...
+                        <option value="7"><%= wishcard.childFirstName %>, you are awesome!</option>
+                        <option value="8">Lots of love and best wishes, <%= wishcard.childFirstName %></option>
+                        <option value="9"><%= wishcard.childFirstName %>, hope you enjoy my gift!</option>
+                        <option value="10">Merry Christmas and a Happy New Year</option>
+                        <option value="11"><%= wishcard.childFirstName %>, have a happy holiday</option>
+                        <option value="12">Congratulations, <%= wishcard.childFirstName %></option>
+                    </select>
+                    <input type="submit" id="submit" class="mt-4" value="Send Message">
+                </form>
+
```
gnorbsl, line 66: I would suggest to generate an array in the backend and just iterate over it here, gives us more flexibilty and cleans up the code. maybe we can add a placeholder in the strings like %name% and replace every occurence of the placeholder with the name.

Not neccessary though

views/wishCardFullPage.ejs
```diff
@@ -40,44 +44,39 @@
...
+                </form>
+                
             </div>
 
             <!-- This part needs to show all users' messages sent to this wishcard 
             wishcard model has 'messages' field which is an objectId array of 'Message'-->
-            <div class="show-msg"> 
-                <% if (wishcard.messages.length === 0) { %>
+            <div class="show-msg">
+                <% if (messages.length === 0) { %>
```
gnorbsl, line 72: miising wishcard.?

gnorbsl: Ah different variables, a bit confusing, whats the difference here between wishcard.messages and messages, maybe a different variable name to clarify?

views/wishCardFullPage.ejs
```diff
@@ -40,44 +44,39 @@
...
                 <h6 class="comf-font">No one signed <%= wishcard.childFirstName %>'s card yet.
                     Be the first person to send a message.</h6>
                 <% } else { %>
                 <div>
-                    <% for (var i = 0; i < wishcard.messages.length; i++) { %>
-                    <p>This line needs to show each user's message sent to this wishcard</p>
-                    <p>Sent by: <%= user.fName %></p>
+                    <% for (var i = 0; i < messages.length; i++) { %>
```
gnorbsl, line 77: I suggest using let instead of var since var is kind of deprecated

gnorbsl (changes requested):

views/wishCardFullPage.ejs
Jacobjeevan: Yeah, probably a good idea to do that. I will change 73 to just messages.

routes/wishCards.js
```diff
@@ -5,206 +5,264 @@ serving all wishcard related routes
...
-        cb(null, false);
-    }
-}
+  // reject a file
+  if (
+    file.mimetype === "image/jpeg" ||
+    file.mimetype === "image/jpg" ||
+    file.mimetype === "image/gif" ||
+    file.mimetype === "image/png"
+  ) {
```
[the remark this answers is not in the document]
Jacobjeevan: Lmao

routes/wishCards.js
Jacobjeevan: Nah, double ticks aren't necessary there :+1:

views/wishCardFullPage.ejs
stacysealky: sorry that was me lol too used to using var

views/wishCardFullPage.ejs
gnorbsl: ah wait you replaced all wishcard.messages with messages?
sorry git is a bit confusing sometimes

gnorbsl:
@Jacobjeevan can you change the name of the bucket var into S3BUCKET with uppercase?

views/wishCardFullPage.ejs
Jacobjeevan: Yeah, no worries :+1:

PR #30, opened 2020-10-02
in progress with curated item list for agency and cleaned up issues

stacysealky:
@gnorbsl @Tokikko @Jacobjeevan Can one of you review my PR? 

chooseItem.ejs is not finished though, still work in progress.


Jacobjeevan (approved):
Looks good :)

PR #41, opened 2020-10-09
Prevent double signup

gnorbsl:


gnorbsl:
Buttons for signup and agency are now disabled until a server response is received

PR #51, opened 2020-10-14
Feature/don 14 captcha

Tokikko:
Added google captcha to signup page. Changed status code from 206 to 400 when a field is missing in submission form


utilities/googleReCaptcha.js
```diff
@@ -0,0 +1,22 @@
+let got = require('got');
```
gnorbsl, line 1: whats got used for here?

views/signup.ejs
```diff
@@ -16,6 +16,13 @@
   <link rel="stylesheet" href="../css/signup.css">
   <!-- jquery -->
   <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
+  <!-- recaptcha -->
+  <script src="https://www.google.com/recaptcha/api.js" async defer></script>
+  <script>
+    function onSubmit(token) {
+      alert("hehe")
```
gnorbsl, line 23: ?

utilities/googleReCaptcha.js
```diff
@@ -0,0 +1,22 @@
...
+// check if captcha token is valid
+// returns boolean 
+async function validateReCaptchaToken(token) {
+    let googleUrl = `https://www.google.com/recaptcha/api/siteverify?secret=${process.env.GOOGLE_CAPTCHA_KEY}&response=${token}`;
+    return (async () => {
+        try {
+            const res = await superagent.post(googleUrl);
+            const responseBody = res.body;
+            if (responseBody.success === false) {
+                return false;
```
gnorbsl, line 13: just return responseBody.success?

gnorbsl (changes requested):
Please check if your master is up2date with the master here, it looks like you are missing quite a few things in your code

utilities/googleReCaptcha.js
Tokikko: Yeah i could, but i like to be more explicit

PR #52, opened 2020-10-14
Verification message Delay

Jacobjeevan:
- Redirect new users -> agency registration (as earlier)
	- Bind user to session
	- Moved logic for sending verf emails outside
- Display message to verify email address (similar to agency Verification email)
- Changed middleware in app.js
	- It checks if user is partner, finds respective agency and binds to response/local storage
- This is a better approach as it allows to remove a bit of redundant code in routes


Jacobjeevan:
> tests are failing now, either you try to fix them or I do but I wont be able to for at least a day

Yea, working on that :D

PR #53, opened 2020-10-14
Don 9 userprofile

Jacobjeevan:
- Changed to findByIdAndUpdate (as a transaction to avoid race conditions in future)
- Slightly changed middleware in app
	- Ensured that both req is updated as well (when changes are made to user/agency)


gnorbsl:
can you pull the latest master and merge it with your project? thats were all the conflicts come from we had a piush to master like 20 min ago

gnorbsl:
nevermind :D

routes/users.js
```diff
@@ -222,6 +221,11 @@ router.post('/signup', signupValidationRules(), validate, async (req, res) => {
   if (candidate) {
     return sendError(res, 409, 'This email is already taken. Try another');
   } else {
+    if (!password) return sendError(res, 206, { message: 'Password missing' });
```
gnorbsl, line 224: this should not be in here

gnorbsl: validation is now handled elsewere and I removed it a while ago

gnorbsl (changes requested):
where do all the view file changes come from?

routes/users.js
Jacobjeevan: :thinking: It was from master hmm..

gnorbsl: if you compare with current master they are not there

PR #56, opened 2020-10-14
Feature/don 14 captcha google

Tokikko:


routes/validations/googleReCaptcha.js
```diff
@@ -0,0 +1,18 @@
...
+
+// check if captcha token is valid
+// returns boolean 
+async function validateReCaptchaToken(token) {
+    let googleUrl = `https://www.google.com/recaptcha/api/siteverify?secret=${process.env.GOOGLE_CAPTCHA_KEY}&response=${token}`;
+    return (async () => {
+        try {
+            const res = await superagent.post(googleUrl);
+            const responseBody = res.body;
+            return responseBody.success;
```
gnorbsl, line 11: you can return res.body.success here

gnorbsl (changes requested):

routes/validations/googleReCaptcha.js
```diff
@@ -0,0 +1,17 @@
+const superagent = require('superagent');
+
+// check if captcha token is valid
+// returns boolean 
+async function validateReCaptchaToken(token) {
+    let googleUrl = `https://www.google.com/recaptcha/api/siteverify?secret=${process.env.GOOGLE_CAPTCHA_KEY}&response=${token}`;
```
gnorbsl, line 6: can you return true if process.env.NODE_ENV is "test"

gnorbsl (changes requested):

PR #57, opened 2020-10-15
Don-17 login without verification

Jacobjeevan:
- Moved email verification to its own function (outside /post/ signup route, so we can reuse it)
	- To do: Add a resend email verification link to unverified profiles
- POST signup
	- Modified return object
	- Removed email (frontend doesn't display the toast for it anymore)
	- Added url (used to redirect to next page - depending on user role)
- Removed check for email verification
	- User can login without verifying their email
- Modified verification messages (since user can login right away, no need to say you can login now :))
- Modified tests

### DIFF
--- a/routes/users.js
+++ b/routes/users.js
@@ -208,6 +208,13 @@ router.post(
   }
 );
 
+const sendEmail = (email, verificationHash) => {
+  sendVerificationEmail(email, verificationHash).then((emailResponse) => {
+    emailResponse = emailResponse ? emailResponse.data : '';
+    if (process.env.NODE_ENV === 'development') console.log(emailResponse);
+  });
+};
+
 // @desc    Create a newUser, hash password, issue session
 // @route   POST '/users/signup'
 // @access  Public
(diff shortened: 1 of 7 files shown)


routes/users.js
```diff
@@ -107,15 +107,14 @@ router.get('/login', redirectProfile, (req, res) => {
 router.get('/profile', redirectLogin, async (req, res) => {
   try {
     let user = req.session.user;
-    let params = { user };
     if (user.userRole === 'partner') {
       let agency = await Agency.findOne({ accountManager: user._id });
+      // If user hadn't filled out agency info, redirect them to form
       if (!agency) {
-        return res.render('agency');
+        return res.status(200).render('agency');
```
gnorbsl, line 114: isnt status 200 default? just out of interest

test/users.test.js
```diff
@@ -335,13 +344,23 @@ describe('Users', () => {
...
+    });
 
-    it('it should not login without verified email', (done) => {
+    it('it should show agency registration page after initial registration for partners', (done) => {
+      signupRequest.userRole = 'partner';
       agent
         .post('/users/signup')
         .send(signupRequest)
         .end((err, res) => {
-          res.should.have.status(200);
```
gnorbsl, line 344: why should it not have 200?

gnorbsl (changes requested):
This will not block users from donating with unverfiied emails or am I missing something?

Jacobjeevan:
> This will not block users from donating with unverfiied emails or am I missing something?

Nah, it won't. Hmm, I can render the donate button based on that though.

test/users.test.js
Jacobjeevan: Oops; accidentally removed that. Adding it back :+1:

routes/users.js
Jacobjeevan: It is; I think I was testing something and added that iirc, not that it makes any difference anyways lol.

gnorbsl:
maybe you can show a message in the modal that they have to verify and disable the button?

Jacobjeevan:
> maybe you can show a message in the modal that they have to verify and disable the button?

Yeah, I added that to the email message in profile :+1: (It just says some features are disabled until email verification).

gnorbsl:
could you add a kind of hover tooltip that shows once you hover over the donate button informing the user that they have to verify their mail? I think you just need to add a title to the button for that

https://stackoverflow.com/questions/2238239/tooltips-for-button

gnorbsl:
ah nvm you are hiding the button completely, it kind of could confuse people

gnorbsl:
I think this will work for now we might have to change that in the future

PR #60, opened 2020-10-16
feat/prettier-eslint

Enubia:
## eslint and prettier
I've gone ahead and added eslint with prettier for code styling with a base config from airbnb
if there are any adoptions you want to make feel free to edit the .eslintrc file

## husky
I also enabled husky which is a commit hook library to lint the code before a commit will be staged to ensure code stylings

## refactor error handler middleware
I refactored the error.js middleware file so it can be used anywhere we would reply a res.status. It also contains a custom Error class which can be used to throw custom errors if needed.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -5,13 +5,12 @@ name: Node.js CI
 
 on:
   push:
-    branches: [ master ]
+    branches: [master]
   pull_request:
-    branches: [ master ]
+    branches: [master]
 
 jobs:
   build:
-
     runs-on: ubuntu-latest
 
     strategy:
@@ -28,7 +27,7 @@ jobs:
       - run: npm ci
       - run: sudo docker run --name mongo -d -p 27017:27017 mongo
       - run: npm run build --if-present
-      - run: npm run format
+      - run: npm run lint
       - run: npm test
       - name: Commit changes
         uses: stefanzweifel/git-auto-commit-action@v4.1.2
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,18 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Attach app",
+      "port": 9229,
+      "request": "attach",
+      "skipFiles": [
+        "<node_internals>/**",
+        "node_modules/**"
+      ],
+      "type": "node",
+      "restart": true,
+      "internalConsoleOptions": "neverOpen",
+      "protocol": "inspector"
+    },
+  ]
+}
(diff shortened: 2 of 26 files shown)


.github/workflows/nodejs.yml
```diff
@@ -28,7 +27,7 @@ jobs:
       - run: npm ci
       - run: sudo docker run --name mongo -d -p 27017:27017 mongo
       - run: npm run build --if-present
-      - run: npm run format
```
gnorbsl, line 31: fromat was autoformatting the code when added to master, lint only checks if the code is formatted and fails if not right?

.vscode/launch.json
```diff
@@ -0,0 +1,18 @@
+{
```
gnorbsl, line 1: this should be in git ignore?

controllers/email.js
```diff
@@ -94,58 +49,85 @@ const getTransport = async () => {
...
 
-  const body = template.replace(
-    'linkplaceholder',
-    `${process.env.BASE_URL}/users/verify/${hash}`
-  );
+  const body = template.replace('linkplaceholder', `${process.env.BASE_URL}/users/verify/${hash}`);
 
   console.log('BODY');
-  return await sendMail(
+  return sendMail(
```
gnorbsl, line 119: wont this return a promise now? not sure if that will cause issues in the signup method

routes/users.js
```diff
@@ -175,47 +163,42 @@ router.get('/agency', redirectLogin, async (req, res) => {
...
+    });
+  } catch (err) {
+    console.log(err);
+    return handleError(res, 400, err);
   }
-);
+});
 
 const sendEmail = (email, verificationHash) => {
   sendVerificationEmail(email, verificationHash).then((emailResponse) => {
```
gnorbsl, line 199: we decided to go with async/await for the project

gnorbsl: not sure if thats already existing code

gnorbsl (changes requested):
looks great thanks for all the cleanup a few minor things though

routes/users.js
gnorbsl: can you reformat this into async/await?

.github/workflows/nodejs.yml
Enubia: lint does formatting as well since I configured the linter script to also fix things :)

.vscode/launch.json
Enubia: can be in gitignore if you wish so, most of the time I see it added so that people who use vscode already have the setup for it

routes/users.js
Enubia: that's exisiting code, I only added the new errorHandler but can refactor it if you wish

.vscode/launch.json
gnorbsl: well it cant hurt I guess

routes/users.js
gnorbsl: yeah it's easier if you just reformat it quickly, thank you

gnorbsl (approved):

PR #61, opened 2020-10-16
added logic for password reset

gnorbsl:


controllers/email.js
```diff
@@ -84,44 +110,39 @@ const sendMail = async (from, to, subject, message, attachments = undefined) =>
...
-      cid: 'Img1_2x.jpg', // same cid value as in the html img src
-    },
-  ];
-
-  const body = template.replace('linkplaceholder', `${process.env.BASE_URL}/users/verify/${hash}`);
-
-  console.log('BODY');
+
+  const body = template
+    .replace('linkplaceholder', `${process.env.BASE_URL}/users/verify/${hash}`)
```
Enubia, line 115: intended to not use ```%```linkplaceholder```%```?

controllers/email.js
```diff
@@ -84,44 +110,39 @@ const sendMail = async (from, to, subject, message, attachments = undefined) =>
...
-    attachments,
+    templateAttachments,
+  );
+};
+
+
+const sendPasswordResetMail = async (to, hash) => {
+
+  const body = template
+    .replace('linkplaceholder', `${process.env.BASE_URL}/users/password/reset/${hash}`)
```
Enubia, line 134: same here as mentioned above

routes/users.js
```diff
@@ -387,4 +392,119 @@ router.get('/choose', redirectLogin, async (req, res) => {
...
+// @access  Private
+// @tested
+router.post('/password/request', async (req, res) => {
+  try {
+
+    if (!req.body.email) return handleError(res, 400, 'email missing');
+
+    const userObject = await User.findOne({ email: req.body.email });
+
+    if (!userObject) return handleError(res, 400, 'user not found');
```
Enubia, line 421: no need to return the handleError, it sends a response inside of it

Enubia (changes requested):
looks good so far, just a few things unlcear

routes/users.js
gnorbsl: if I dont return the rest of the code is getting called though?

PR #69, opened 2020-10-16
refactor: new project structure, db repositories

Enubia:


listtree.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
gnorbsl, line 1: I assume this is a leftover from your structure template?

nodemon.json
```diff
@@ -1,11 +1,7 @@
...
-    "controllers",
-    "middleware",
-    "models",
-    "routes",
-    "utilities"
+    "server"
   ],
   "ext": "js",
-  "exec": "node --inspect=0.0.0.0:9229 app.js"
+  "exec": "NODE_ENV=development node --inspect=0.0.0.0:9229 server/app.js"
```
gnorbsl, line 6: this might break takikkos setup, NODE_ENV dev should be filled automatically on startup by dotenv

projectStructure.md
```diff
@@ -0,0 +1,28 @@
+## Project structure
```
gnorbsl, line 1: I think this can be removed?

stacysealky:
Wow, thanks for cleaning! it was much needed. 
Just make sure that we are linking all static assets from the changed directories, etc.  
I haven't fully reviewed this, but I think the chooseItem.ejs and defaultItems.js use all local img assets. So make sure they are still linked please. 

@gnorbsl please review this PR thanks

server/routes/users.js
```diff
@@ -52,7 +54,7 @@ const redirectProfile = (req, res, next) => {
 // @tested 	Yes
 router.get('/', (req, res) => {
   try {
-    res.status(200).render('home', {
+    res.status(200).render(path.join(__dirname, '../../client/views/home'), {
```
gnorbsl, line 57: those are not neccessary, we just have to tell ejs where the views are, itll get the correct ones then

gnorbsl (changes requested):

gnorbsl (approved):

gnorbsl:
lets hope I wont regret this

PR #70, opened 2020-10-17
fix: only link in button clickable

Enubia:


stacysealky:
oh boy not the linter issue again lol

stacysealky:
looks like you only changed one thing? wrapping the <button> with <a> links, right?

Enubia:
yes only the order of those tags were messed up

PR #72, opened 2020-10-18
feat: google and facebook login setup

Enubia:
- [x] add google/fb login buttons
- [x] create routing for both

~~handle logout~~ no logout needed since it would also logout the user from every other open session

~~we still need a facebook test account and need to enable the local environments in google dashboard so that we can proceed with the development @stacysealky~~ this is done


nodemon.json
```diff
@@ -3,6 +3,6 @@
     "server",
     "client"
   ],
-  "ext": "js, ejs",
+  "ext": "js",
```
Enubia, line 6: @gnorbsl no need to restart the server on ejs file changes

gnorbsl: Ah makes sense, I somehow thought that pages get rendered once on server start :D

Enubia:
@gnorbsl @stacysealky please review

nodemon.json
Enubia: nope, only on load in browser, we don't use webpack yet for hot reloading

stacysealky:
@Enubia This looks great. Reviewed 16/20 files. @gnorbsl can you review the 4 files that aren't reviewed yet?

server/db/repository/UserRepository.js
```diff
@@ -3,31 +3,31 @@ const User = require('../models/User');
 
 async function getUserByObjectId(id) {
   try {
-    return User.findOne({ _id: id });
+    return User.findOne({ _id: id }).exec();
```
gnorbsl, line 6: why the exec now?

Enubia: https://mongoosejs.com/docs/promises.html#queries-are-not-promises

PR #84, opened 2020-10-21
Don 40 wishcard bug

Jacobjeevan:
- Simple fix for the template tag


Jacobjeevan:
Closing it for now - I think I need to update with master again.

PR #86, opened 2020-10-21
feat: added slack integration

Enubia:
we now  use slack instead of email for feedback form

### DIFF
--- a/server/routes/about.js
+++ b/server/routes/about.js
@@ -11,7 +11,7 @@ const router = express.Router();
 // IMPORT WISHCARD MODEL
 const ContactRepository = require('../db/repository/ContactRepository');
 // LOAD EMAIL SENDING FUNCTION
-const { sendMail } = require('../controller/email');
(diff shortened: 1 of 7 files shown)


Enubia:
@stacysealky @gnorbsl

server/routes/about.js
```diff
@@ -59,4 +59,17 @@ router.post('/email', async (req, res) => {
   }
 });
 
+router.post('/customer-service', async (req, res) => {
+  const { name, email, subject, message } = req.body;
+  const done = await sendSlackFeedbackMessage(name, email, subject, message);
+
+  if (done) {
+    return res.status(200).send({
```
stacysealky, line 67: we should do toasts for all success msgs in the future? @gnorbsl @Enubia @Tokikko - did someone already work on this?

stacysealky (commented):
👍

server/routes/about.js
Enubia: We could,maybe just make the toast thing configurable so that I can show different colors based on what you want to show

PR #94, opened 2020-10-21
Don 46/toast error

Enubia:
- [x] added error middleware with 500 page
- [x] added proper field validation for toast messages
- [x] added morgan request logger for dev environment
~~replaced old logger with winston as logger~~
- [x] replaced winston with log4js due to performance reasons


gnorbsl:
Dont use winston, winston will break under load. its a 3 year old problem that they never fixed.

I suggest logger4js

Enubia:
@stacysealky @gnorbsl please review :rocket:

gnorbsl (approved):

PR #97, opened 2020-10-22
feat: added request key validation

Enubia:
removed cors since it was added by accident
added option for toastify to display success toast (green color)


gnorbsl:
this is on your sanitation branch but isnt sanitizing anything?

gnorbsl (approved):

PR #103, opened 2020-10-23
Don 5 wishcard tests

Jacobjeevan:
- Added wishcard tests
- Some other changes were made to make testing smoother
- Also wrapped logic for multer/aws setup in a check for environment (only setup for production; I had used one of my buckets initially, but I think this is a better setup than adding S3/AWS credentials into test env file).


server/routes/wishCards.js
```diff
@@ -27,10 +42,26 @@ const {
...
-const s3 = new AWS.S3({
-  accessKeyId: process.env.AWS_KEY,
-  secretAccessKey: process.env.AWS_SECRET,
-});
+const { getMessageChoices } = require('../utils/defaultMessages');
+
+let s3;
+let s3storage;
+
+if (process.env === 'production') {
```
gnorbsl, line 50: can you change the check to LOCAL_DEVELOPMENT so that s3 is still used on the dev server

gnorbsl: aside from that process.env will never be production, process.env.NODE_ENV

gnorbsl: for that we have USE_AWS too, so this one is not really neccessary anyway

server/routes/wishCards.js
```diff
@@ -80,89 +102,102 @@ const AgencyRepository = require('../db/repository/AgencyRepository');
...
-        childBirthday: new Date(req.body.childBirthday),
-        wishItemPrice: Number(req.body.wishItemPrice),
-        wishCardImage: req.file.location,
-        createdBy: res.locals.user._id,
-        // Uncomment once address fields are added to profile page.
-        /* address: {
+      });
+    } else {
+      try {
+        let newWishCard = await WishCardRepository.createNewWishCard({
```
gnorbsl, line 122: why let if the variable is never changed again?

server/helper/wishcardHelper.js
```diff
@@ -0,0 +1,54 @@
...
+  if (!user) {
+    res.status(403).redirect('/users/login');
+  } else if (await IfNullOrNotVerifiedAgency(userInfo)) {
+    res.status(403).redirect('/users/profile');
+  } else {
+    next();
+  }
+};
+
+const renderPermissions = async (req, res, next) => {
```
gnorbsl, line 39: code is almost identical; im sure there is a way to improve that

server/routes/wishCards.js
```diff
@@ -80,89 +102,102 @@ const AgencyRepository = require('../db/repository/AgencyRepository');
...
+  },
+);
 
 // @desc    Render wishCards.html which will show all wishcards
 // @route   GET '/wishcards'
-// @access  Public, all users can see
+// @access  Private, all users can see
 // @tested 	Yes
-router.get('/', async (_req, res) => {
+router.get('/', redirectLogin, async (_req, res) => {
```
gnorbsl, line 200: doesnt this prevent everyone from seeing wishcards who is not logged in

gnorbsl (changes requested):

server/routes/wishCards.js
Jacobjeevan: Changed it to const :+1:

server/routes/wishCards.js
Jacobjeevan: Yeah, true. Changed that back; changed tests as well.

server/helper/wishcardHelper.js
Jacobjeevan: Don't have any ideas as of now; any pointers? :thinking:

Enubia:
```js
const permissionHandler = (redirect = false) => {
  return (req, res, next) => {
    const { user, agency } = req.session;
    const userInfo = { user, agency };
    if (redirect) {
      if (!user) {
        res.status(403).redirect('/users/login');
      } else if (await IfNullOrNotVerifiedAgency(userInfo)) {
        res.status(403).redirect('/users/profile');
      } else {
        next();
      }
    } else {
      if (!user) {
        res.status(403).send({ success: false, url: '/users/login' });
      } else if (await IfNullOrNotVerifiedAgency(userInfo)) {
        res.status(403).send({ success: false, url: '/users/profile' });
      } else {
        next();
      }
    }
  }
}
```

PR #109, opened 2020-10-25
Feature/don 56 admmin draf wishcards

Tokikko:


server/routes/wishCards.js
```diff
@@ -221,6 +222,59 @@ router.get('/me', async (req, res) => {
...
+      const wishCardId = mongoSanitize.sanitize(req.body.wishCardId);
+      await WishCardRepository.updateWishCardStatus(wishCardId);
+      return res.status(200).send({
+        success: true,
+        error: null,
+        data: null,
+      });
+    } catch (error) {
+        return res.status(200).send({
+            success: false,
```
Enubia, line 271: You can use the 
```js 
handleError(res, 200, error);
``` 
function here from helper/error.js since that one also logs the error

Enubia: also, why are we return 200 on failure?

server/db/repository/WishCardRepository.js
```diff
@@ -55,6 +63,14 @@ async function pushNewWishCardMessage(id, message) {
   }
 }
 
+async function updateWishCardStatus(id) {
+    try {
```
Enubia, line 67: should be named accordingly to what it does, in this case 
``` setWishCardPublished```
or better would be to pass the status as parameter so we can also use this to unpublsih a wishcard

```js
async function updateWishCardStatus(id, status) {
    try {
      return WishCard.updateOne({ _id: id }, {$set: { status }}).exec();
    } catch (error) {
      throw new Error(`Failed to update Wishcard messages: ${error}`);
    }
}
```

Enubia (changes requested):

server/routes/wishCards.js
Tokikko: A mistake :)

server/db/repository/WishCardRepository.js
Tokikko: Well it updates a value, the function is also called updateone but i have renamed it to setWishCardStatus

Enubia: Sorry what I meant was if you keep the name than add the status parameter so that it can be reused and if not I would rename it

gnorbsl (approved):

PR #117, opened 2020-10-26
Agency organization

Tokikko:
Added agency address info


gnorbsl:
can you add validations for the new fields?

Tokikko:
> 
> 
> can you add validations for the new fields?



> 
> 
> can you add validations for the new fields?

Yeah, will do

server/routes/validations/users.validations.js
```diff
@@ -42,16 +42,40 @@ const updateProfileValidationRules = () => {
...
   return [
-    body('agencyName').notEmpty().isString().withMessage('Agency name is required!'),
+    body('agencyName')
+      .notEmpty()
+      .isString()
+      .withMessage('Agency name is required!'),
     body('agencyWebsite').optional(),
+    body('agencyAddress.address1')
+      .notEmpty()
+      .isLength({min: 5, max: undefined})
```
gnorbsl, line 52: you can leave max out I think, itll default to undefined

server/routes/validations/users.validations.js
```diff
@@ -42,16 +42,40 @@ const updateProfileValidationRules = () => {
...
+      .withMessage('Address must containt at least 5 characters'),
```
gnorbsl, line 53: typo

server/routes/validations/users.validations.js
```diff
@@ -42,16 +42,40 @@ const updateProfileValidationRules = () => {
...
+      .isLength({min: 2, max: undefined})
+      .withMessage('State must containt at least 2 characters'),
+    body('agencyAddress.country')
+      .notEmpty()
+      .isLength({min: 2, max: undefined})
+      .withMessage('Country must containt at least 2 characters'),
+    body('agencyAddress.zipcode')
+      .notEmpty()
+      .isLength({min: 5, max: undefined})
+      .withMessage('Zipcode must containt at least 5 characters'),
```
gnorbsl, line 70: TYPOS EVERYWHERE!

server/routes/validations/users.validations.js
```diff
@@ -42,16 +42,40 @@ const updateProfileValidationRules = () => {
...
     body('agencyPhone')
       .notEmpty()
-      .isNumeric()
       .isLength({ min: 7, max: undefined })
       .withMessage('Phone number must be at least 7 characters long')
-      .matches(/^(?=\d).{7,}$/)
-      .withMessage('Phone number must contain only numbers'),
+      .matches(/^\(?([0-9]{3})\)?([ .-]?)([0-9]{3})\2([0-9]{4})/) 
+      .withMessage('Phone number must match format XXX-XXX-XXXX'),
```
gnorbsl, line 76: this might cause issues, no one writes their phone number like this, better only check if phonenumber is an integer for now

gnorbsl (changes requested):

server/routes/validations/users.validations.js
Tokikko: Stacy specifically asked for this:
https://donategifts.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=DON&modal=detail&selectedIssue=DON-63

server/routes/validations/users.validations.js
Tokikko: When you copy a typo it makes more typos everywhere

gnorbsl (approved):

server/routes/validations/users.validations.js
stacysealky: So I've just noticed that people type phone numbers in several different ways. 

- 777-777-7777
- 7777777777
- (+1) 777 777 7777
- (777) 777-7777
- 777.777.7777

I mean this can go on, but you get the point though. 
So how do we validate it without making people type again and again to fit our format? 
Suggestions please

gnorbsl: I would say we go with only numbers and check the phone number on every key down event, or we filter out all non numbers on submit

PR #123, opened 2020-10-27
Don 57 error messages

Jacobjeevan:
-  Wishcard creation/error message fix in ChooseItem
- Wrapped error messages in an object (with 'msg' as a parameter)
	- Allows for consistency (we were already using msg parameter in some cases)
	- Modified ClientJS code slightly to ensure msg is parsed and passed to Toast


client/public/clientJS/passwordReset.js
```diff
@@ -21,7 +21,9 @@ $(document).ready(function () {
             },
             error: function(response, textStatus, errorThrown) {
                 $("#submit-btn").prop("disabled",false);
-                showToast(response.responseJSON.err);
+                let txtToJson = response.responseJSON;
+                let { msg } = txtToJson.error;
+                showToast(msg);
```
gnorbsl, line 26: you can probably just do response.responseJSON.error.msg?

gnorbsl (changes requested):
I think it makes more sense to put the "msg" into the handleError function instead of changing it 100 times everywhere else

gnorbsl:
And I'm not sure how useful it is to add a msg field in general since the error has no other data attached to it has it?

Jacobjeevan:
> I think it makes more sense to put the "msg" into the handleError function instead of changing it 100 times everywhere else

> And I'm not sure how useful it is to add a msg field in general since the error has no other data attached to it has it?

Well, alternative was to have conditionals on client side (since some of the errors are objects - like validation for instance, while others are just strings). But, good point about moving it to HandleError :thinking:

client/public/clientJS/passwordReset.js
Jacobjeevan: Yea, this looked bit more nice xD

gnorbsl:
still have "msg" in routes I assume its not changed yet? just saw that you pushed fixes

Jacobjeevan:
> still have "msg" in routes I assume its not changed yet? just saw that you pushed fixes

Should be good now :+1:

gnorbsl:
there is still the problem with the errorhandler, please check my open comment here

gnorbsl (approved):

PR #124, opened 2020-10-27
Socket fix

gnorbsl:
hopefulyl fixing cors issues


stacysealky (approved):

server/app.js
```diff
@@ -42,6 +43,7 @@ if (process.env.NODE_ENV === 'development') {
   // colorful output for dev environment
   app.use(morgan('dev'));
 }
+app.use(cors());
```
Enubia, line 46: I would add cors only to the routes where you need it, with this every route is accessible from outside, you can add it as middleware on specific routes if that is needed for sockets

server/routes/wishCards.js
```diff
@@ -227,48 +237,48 @@ router.get('/me', async (req, res) => {
...
-        }
-      });
-    } catch (error) {
-      handleError(res, 400, error);
+  try {
+    const WISHCARD_STATUS = "draft";
+    const USER_ROLE = "admin";
+    // only admin users can get access
+    if (res.locals.user.userRole !== USER_ROLE) {
+      return res.status(401).render('401');
```
Enubia, line 245: How about rendering 404? Doesn't disclose that we have admin specific pages

server/routes/wishCards.js
```diff
@@ -227,48 +237,48 @@ router.get('/me', async (req, res) => {
...
-        error: null,
-        data: null,
-      });
-    } catch (error) {
-        handleError(res, 400, error);
+  try {
+    const USER_ROLE = "admin";
+    // only admin users can get access
+    if (res.locals.user.userRole !== USER_ROLE) {
+      return res.status(401).render('401');
```
Enubia, line 270: Same here with the 404 proposal

server/routes/wishCards.js
```diff
@@ -411,48 +421,198 @@ router.post(
...
         }
 
-        const lockedWishCard = await WishCardRepository.lockWishCard(wishCardId, user._id);
-        io.emit('block', { id: wishCardId, lockedUntil: lockedWishCard.isLockedUntil });
+        if(isDonated) {
+
+          const wishCard = await WishCardRepository.getWishCardByObjectId(wishCardId);
+
+          wishCard.isDonated = true;
+          wishCard.save();
```
Enubia, line 580: Wishcard isn't used anywhere else from here on, would move the updating to the repository and only make the function call to update it, makes it reusable

Enubia (changes requested):

PR #133, opened 2020-10-29
seeder data for dev server

MarkellRichards:
Stock seeder data for wish cards and users.

### DIFF
--- a/server/db/seeder.js
+++ b/server/db/seeder.js
@@ -1,0 +1,44 @@
+const path = require('path');
+const { users, wishcards } = require('./seederData');
+
+const User = require('./models/User');
+const WishCard = require('./models/WishCard');
+
+const MongooseConnection = require('./connection');
+
+require('dotenv').config({ path: path.resolve(__dirname, '../../config/config.env') });
+
+MongooseConnection.connect();
+
+const importData = async () => {
+  try {
+    await User.deleteMany();
+    await WishCard.deleteMany();
+    await User.insertMany(users);
+    await WishCard.insertMany(wishcards);
+
+    // console.log('Data imported!');
+    process.exit();
+  } catch (error) {
+    // console.error(error);
+    process.exit(1);
+  }
+};
+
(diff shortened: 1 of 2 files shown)


gnorbsl:
looks good so far, can you add an agency to the user that is partner, and make sure that the script doesnt run and exists with error if process.env.LOCAL_DEVELOPMENT is false

server/db/seeder.js
```diff
@@ -0,0 +1,38 @@
...
+    await WishCard.deleteMany();
+    process.exit();
+  } catch (error) {
+    process.exit(1);
+  }
+};
+if (process.argv[2] === '-d') {
+  destroyData();
+} else {
+  importData();
```
Enubia, line 37: would also be nice to check for -i, so that we can insert only, a few of us already have data in their local db

MarkellRichards:
> looks good so far, can you add an agency to the user that is partner, and make sure that the script doesnt run and exits with error if process.env.LOCAL_DEVELOPMENT is false

I am looking at the User model. There is no reference to the agency model.

server/db/seeder.js
MarkellRichards: Got it, will add option

Enubia:
IMO the agency should hold all the information about users and wishcards related to them since it makes the most sence.

Users -> Agency <- Wishcards

maybe we should discuss this

gnorbsl:
The agency model holds a userId in accountmanager @MarkellRichards , thats what I meant

Enubia:
@MarkellRichards I would take the models and fill all the fields that are available, copy them like 20 times and change some of the data, 3 samples per collection aren't really that much for testing

Enubia:
@MarkellRichards are you done with the seeder data now? or is there anything you want to add to it?

MarkellRichards:
I just need to add more sample data. In the middle of traveling so I won't have internet till tomorrow morning.

PR #136, opened 2020-10-30
DON-24: Create login modal, needs styling

mariawhen:
- Creates login modal
- Triggered on donate button if user is not logged in
- Didn't style the oauth buttons
- ignore Last pass icons within input fields

`! Important: this needs testing. I couldn't get local wishcards populated..`

### Desktop View
Screen Shot 2020-11-01 at 10 25 55 PM

### Mobile View
Screen Shot 2020-11-01 at 10 42 38 PM

### DIFF
--- a/client/public/css/home.css
+++ b/client/public/css/home.css
@@ -527,3 +527,85 @@ New color scheme reference is in utils.css
   text-align: center;
   background: rgba(49, 49, 49, 1);
 }
+
+/*--------------------------------------------------------------
+    Login Modal
+--------------------------------------------------------------*/
+.modal.login .modal-header,
+.modal.login .modal-footer {
+  border: none;
+}
+
+.modal.login .modal-content {
+  border-radius: 20px;
+}
+
+.modal.login h4 {
+  text-align: center;
+  margin-top: 20px;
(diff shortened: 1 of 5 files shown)


client/public/css/home.css
```diff
@@ -527,3 +527,92 @@ New color scheme reference is in utils.css
...
+}
+
+.modal.login .forgot-container {
+  text-align: right;
+  width: 100%;
+  margin-bottom: 20px;
+}
+
+.modal.login a {
+  color: rgba(0, 161, 154, 1) !important;
```
Enubia, line 572: Please avoid important if possible, use a more specific selector in that case

Enubia (changes requested):

client/public/css/home.css
mariawhen: There's already an important coming from utils.css `color: inherit !important;`

I can't even override with an id selector unless I use important

Enubia: Ok ya just checked it, well in that case than it's ok, but for the future we should avoid them :D

Enubia (approved):

PR #139, opened 2020-10-30
Checkbox for agency address

Tokikko:
Added checkbox on wishcard creation. If users select it it will prefill wishcard address form with agency address

### DIFF
--- a/client/public/clientJS/profile.js
+++ b/client/public/clientJS/profile.js
@@ -58,6 +58,65 @@ $(document).ready(() => {
 // About me edit. Click handler
 window.onload = function () {
   let buttonSaveProfile = document.getElementById('save-about-me-button');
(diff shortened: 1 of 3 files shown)


client/public/clientJS/profile.js
```diff
@@ -58,6 +58,65 @@ $(document).ready(() => {
...
+          else {
+            showToast("Agency address not present. Most likely that information is missing, contact administrator");
+          }
+        })
+        .catch((error) => {
+          showToast("Could not get agency address");
+        })
+      }
+      else {
+        showToast("Address form must be empty if we want to use agency address so that there is no accidental overwrite");
```
gnorbsl, line 87: the toast might not be shown long enough for errors like this I think. Is there a very to just ask the user to confirm if they want to overwrite the data?

gnorbsl (approved):

PR #140, opened 2020-10-30
Feat/don 74 wishcardsorting

Enubia:
## INFO

whats not included here is the most recent search option since I think that wouldn't be fair

one other thing is the including of non active cards:
i'm not including them in the result since i don't see the real value for this, instead you can set a filter to show only already donated cards


gnorbsl:
jeez are you trying to get the employee of the month award :D

gnorbsl (approved):

PR #141, opened 2020-10-30
scss introduction

MarkellRichards:
Organized css files into the 7-1 architecture. Added new watch:all script so nodemon and watch:sass run at same time. Wishcard designs were reverted and UX was added.


stacysealky:
don't merge this until v1.0 deployed -- too many conflicts, low priority

PR #155, opened 2020-11-06
Don 29 how to page / profile page responsiveness

MarkellRichards:
Finished how-to page, (Stacey will add dynamic content based on select value). Cleaned up the profile page for v1. Added SVG's for how-to page and some custom styling.


stacysealky (approved):
Looks good. Clean css. muy bien

PR #156, opened 2020-11-06
refactor: downloaded dependencies

Enubia:
mainly bootstrap js and css as well as jquery and popper
we are serving them statically now so that the initial page load is faster
with this the user doesn't need to download it anymore from a cdn
also updated the button in wishcarddetail page

# package-lock.json changes
since i am using npm@7 it created a new lock file called v2
i would encourage you guys to upgrade to npm@7
since it loads dependencies way faster and also automatically installs
peer deps


MarkellRichards (approved):
Looks good to me

MarkellRichards (approved):
Everything checks out

gnorbsl (approved):

PR #160, opened 2020-11-08
Login Toast Responsive for phones

deep-a:
Added login toast message to be responsive for phones plus new CSS styles.


MarkellRichards (commented):
Can you try using BEM for your class names. Such as login__link. When I revise our project to scss. I will be converting a lot of class names. 

To clarify the login is the block, link is the element, and currently there's no modifier such as a different color.

deep-a:
Will reopen once Fork in synced.

phffmn:
@deep-a can you check if we can add this:

https://stackoverflow.com/questions/32155036/show-jquery-popup-only-at-first-visit-of-a-page

to only show the toast at first visit

stacysealky (approved):

PR #168, opened 2020-11-11
WishCard fixes

Jacobjeevan:
- /users/agency/wishcards -> /wishcards/me

Both routes were present, but only the former was used for functionality; I moved the logic for former into latter (which is now used for rendering agencyWishcards aka _My Wishcards_ page.

Also added logic to agencyWishcards for displaying wishcards based on status (Need to be beautified further; as well as add edit/delete functionality to draft wishcards).

- Bug fix

Wishcards created using guided route were not being pushed to their respective agencies. Added logic for that. POST routes were also pushing wishcards directly, rather than the id, so fixed that as well. We might have to look into production, if any agencies have used guided route for Wishcard creation (luckily, we do have the createdBy field to go on, but we will have to fix it).

- Removed wishcards array in User model

It wasn't used anywhere (and wishcards weren't pushed to user anyways).

### DIFF
--- a/client/views/agencyWishCards.ejs
+++ b/client/views/agencyWishCards.ejs
@@ -1,71 +1,171 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <title>See Wish Cards | Donate Gifts</title>
-    <!-- ** NOTE: CDN Bootstrap.js, Popper.js, and jQuery are moved to the bottom so the page loads faster-->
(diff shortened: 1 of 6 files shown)


stacysealky:
@Jacobjeevan 'wishcards' in user model was for later feature, which will be used for saving/ starring the wishcard and will be able to view them in user/profile

stacysealky (approved):
looks good

PR #169, opened 2020-11-11
Slight touch up on carousels

MarkellRichards:
Home cards are no longer cut off. Fixed shadows for all cards for consistency. Altered arrows for both home carousels to look the same.


stacysealky (approved):
looks good, but can we all be in sync with the indentation changes? lots of unnecessary deletion/addition happening here

PR #171, opened 2020-11-11
Don 126 login modal

Jacobjeevan:
LoginModal Adjustments

- LoginModal reloads the page, rather than redirecting.
- Few other adjustments needed to be made to facilitate this, since /login route was redirecting on success and rendering login page (with errorMessages) on failure.
- Now, it will simply send an object in either case (with url for redirection and errorMessages - both of which are handled in the client side).

Update: ~~Also fixed a minor error with messages (redirectLogin rather than renderPermissions. Latter limited messaging to only agency users - which was a bug).~~

Some changes need to be made for tests as well, keeping that for another ticket since it's not related. Originally had kept it with this since it was a one line change.

### DIFF
--- a/client/public/clientJS/loginModal.js
+++ b/client/public/clientJS/loginModal.js
@@ -16,7 +16,7 @@ $(document).ready(function () {
       data: formData,
       success: function (response, textStatus, jqXHR) {
         $('#submit-btn').prop('disabled', false);
-        location.assign(response.url);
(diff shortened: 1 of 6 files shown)


Jacobjeevan:
I have to modify the tests a bit, will do that and push.

stacysealky (approved):
looks good!

PR #189, opened 2020-11-15
Test for logging to kibana

gnorbsl:


gnorbsl:
this is basically a test to log to kibana to make logs more accessible to others and also get juice statistics in a pretty way

Enubia (approved):

PR #192, opened 2020-11-16
Add redirect for live and dev after password reset

Tokikko:


client/public/clientJS/passwordReset.js
```diff
@@ -49,7 +49,13 @@ function getHostName() {
             return "http://localhost:8081";
```
gnorbsl, line 49: I think we can just use process.env.BASE_URL here which will always point to the right domain

gnorbsl (changes requested):

client/public/clientJS/passwordReset.js
Tokikko: I can not use that in frontend though.

gnorbsl (approved):

PR #200, opened 2020-11-17
Disabled button

Tokikko:
Disabled donate button when clicking on read more


client/views/wishCardFullPage.ejs
```diff
@@ -22,15 +22,15 @@
         <div class="section-title mt-5">
           <h1 class="card-title text-center crayon-font grn-font"><%= wishcard.childFirstName %>'s Wishcard</h1>
           <img id="img-kid" src="<%= wishcard.wishCardImage %>" alt="Card image" />
-          <a
+          <!--<a
```
gnorbsl, line 25: probably would be better to actually hide it with ejs like <% if (false) { %> this way its not in the source

Tokikko: I can but kinda weird to have if (false) in code :D

PR #210, opened 2020-11-19
Don 141 unverified donations v2

Jacobjeevan:
- Wishcard routes
	- Added checkVerifiedUser validation for donation routes (card lock/unlock)
- WishCard
	- Replaced isLoggedIn with emailVerified
- WishCardModal
      - Moved scraping logic to their own functions (simulateScrape vs Scrape)
      - Added timeout for simulating a scrape for local dev
	  - helps with mocking UX

v2

**WishCardModal**

- Changed UX Flow
	- If user clicks the button (I checked out...for scraping)
		- If confirmed: Stop and hide countdown and display confirmation message, unlocks card
		- If not confirmed: Change checked out button text
			- Allow user to check again
			- Display message (link to contact us page)
	- If not (countdown gets to 0)
		- Auto scrape
		- If confirmed: display message, unlocks card
		- If not confirmed: Change checked out button text to try again
			- Allow user to check again
			- Display message (link to contact us page)
	- Slight fix to ensure close button is only shown before countdown and after scraping (prevent user from reserving and closing the window early)
- Show/hide status
      - Hide status when user retries scraping (through showLoadingView function)
      - Show status only when signal (donated or not donated) is received from backend


client/public/clientJS/wishCardModal.js
```diff
@@ -27,212 +27,208 @@ $('#wishCardDonateModal').on('show.bs.modal', function (event) {
...
-                    });
+                        <p class="donate-false" id="donateNotDone-${wishCardId}">Changed your mind? Click here to exit.</p>`);
 
+          $('.donate-false').on('click', (event) => {
+            $.ajax({
+              type: 'POST',
+              url: '/wishcards/unlock/' + wishCardId,
+              data: {},
+              success: (response, textStatus, jqXHR) => {
+                console.log(response);
```
Enubia, line 72: please remove all console.logs :+1:

server/routes/wishCards.js
```diff
@@ -554,85 +554,96 @@ router.get('/status/:id', async (req, res) => {
...
-        done(false);
-        return true;
+  try {
+    const wishListArray = /registryID\.1=(.*?)&.*?registryItemID.1=(.*?)&/gm.exec(url);
+    if (wishListArray) {
+      const wishListId = wishListArray[1];
+      const itemId = wishListArray[2];
+      const wishcardInfo = { wishCardId, userId, price, itemId, wishListId };
+      if (wishListId) {
+        if (process.env.LOCAL_DEVELOPMENT) {
```
Enubia, line 642: this should be checked for the testing environment as well
@gnorbsl please correct me if I'm wrong

Enubia (changes requested):

server/routes/wishCards.js
Jacobjeevan: Local dev is for both test and dev, I believe :thinking:

Enubia (approved):

PR #222, opened 2020-11-25
Disable stripe button immediatelly after click

Tokikko:


client/public/clientJS/stripe.js
```diff
@@ -53,6 +56,7 @@ form.addEventListener('submit', function (event) {
         .catch((error) => {
           console.error('Error:', error);
           showToast('An error has occured while processing payment');
+          loading(false);
```
gnorbsl, line 59: do we need to add aother loading(false) call in `payWithCard`?

gnorbsl (changes requested):

client/public/clientJS/stripe.js
Tokikko: Yes, if it fails on our backend we need it to re enable the pay button again

gnorbsl (approved):

PR #227, opened 2020-11-25
Don 151 DonationHistory

Jacobjeevan:
- Donation Schema Changes
- DonationHistory Page (Donations given for donors, Donations received for partners) - possibly beautify this more!
- Donation is saved in Stripe webhook route


server/routes/stripe.js
```diff
@@ -73,6 +74,13 @@ router.post('/webhook', bodyParser.raw({ type: 'application/json' }), async (req
           agency: event.data.object.metadata.agencyName,
         });
 
+        await createNewDonation({
```
gnorbsl, line 77: this doesnt follow the rest of the project i think, everywhere else is repository.function()

while this might be shorter I think its better to keep the same style everywhere.

gnorbsl: also since you dont need a response I dont think you need an await here

server/routes/users.js
```diff
@@ -590,4 +591,26 @@ router.get('/agency/address', async (req, res) => {
 },
 );
 
+router.get("/profile/donations", redirectLogin, async (req, res) => {
+  try {
+    const { user } = req.session;
+    let Donations;
```
gnorbsl, line 597: pls no uppercase variables unless they are classes or similar

gnorbsl (changes requested):

server/routes/users.js
Jacobjeevan: alright :+1:

Jacobjeevan:
Merged #230 into this PR.

client/views/donationHistory.ejs
```diff
@@ -0,0 +1,76 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+    <title>See Donation History | Donate Gifts</title>
+    <!-- ** NOTE: CDN Bootstrap.js, Popper.js, and jQuery are moved to the bottom so the page loads faster-->
+    <%- include('./templates/header')-%>
+    <link rel="stylesheet" href="/public/css/search.css" />
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/socket.io/2.3.1/socket.io.js"
```
gnorbsl, line 9: exists in header now, cdn can be removed

gnorbsl (changes requested):

gnorbsl (approved):

PR #230, opened 2020-11-26
Schema changes and Seeder fix

Jacobjeevan:
**Only merge after #227 since seeder relies on Donation Changes**

**Agency Model**

	- Removed joinedBy field (redundant - model already has a createdBy)
		- This field was not used elsewhere in the project (search found no other match, so safe to remove)

**User model**

	- savedWishcard
		- Will add this back when we get around to implementing this functionality
	- added enum to userRole for validation (ensure role belongs to one of the classes)
	- removed donationsMade (redundant since we use donations Model, which keeps track of donor user as well)

**Messages model**

	- Removed messages array from Wishcard schema
		- Messages already have a messageTo field (removing redundancy)
	- Added getMessagesByWishCardId fn to MessageRepository
		- Wishcard By Id route now uses this to fetch all messages associated with a wishcard
		- Removed earlier method of populating messages as a result
	- Slight change to wishCardFullPage view to display User name
	- Updated seeder accordingly

**Seeder**

	- Added all the related fields to wishCards
		- Push wishcards to agency
	- Added logic for both Donation and Message creation
	- Slight simplification to data in seederData (fewer users)


gnorbsl (approved):

gnorbsl:
is this tested with the current database or only with a clean one?

Jacobjeevan:
> is this tested with the current database or only with a clean one?

:thinking: Running the seeder again (without any additional parameters) will clear out current DB and create new entries (which is what we want since current DB is missing a few relations between the different collections).

gnorbsl:
> > is this tested with the current database or only with a clean one?
> 
> 🤔 Running the seeder again (without any additional parameters) will clear out current DB and create new entries (which is what we want since current DB is missing a few relations between the different collections).

but we cant do that on prod, so how owuld this update affect an existing database

Jacobjeevan:
Merged into #227

PR #233, opened 2020-11-29
checkout page almost done

stacysealky:
- fixed responsiveness
- added 4 helper modals
- fixed frontend layout
- donation add-on function is missing


server/routes/wishCards.js
```diff
@@ -405,11 +405,17 @@ router.get('/donate/:id', redirectLogin, getByIdValidationRules(), redirectLogin
...
+    //and if we are still using agency array, the matching obj is the 1st one
+    //but who knows ¯\_(ツ)_/¯
+  
     res.status(200).render('donate', {
       user: res.locals.user,
       wishcard: wishcard || [],
       extendedPaymentInfo,
-      agencyName: agency[1].agencyName,
+      agencyName: agency[0].agencyName,
+      agency: agency[2]
```
Tokikko, line 418: Is this now compatible with the new seeder? There was a difference in data on DEV and LOCAL database and since im always testing on DEV this worked for me but some people where having problems with local [[agency[1].agencyName]]. Jacob seems to have fixed the seeder but someone with a local instance should test it.

Also if we require the entire agency object in ejs then agencyName is redundant since we can get it from agency itself

PR #236, opened 2020-11-29
DB Migration

gnorbsl:
- Changed wishcardTo field to belongsTo to avoid confusion
- Added migrationsscript that fills belongsTo on existing databases


Jacobjeevan (approved):
:+1: While you are at it, do you also want to take care of the status field as well?

Ensure all existing wishcards have a status of published on prod (I think Stacey was manually setting this, but in case she missed any). 

We can then revert GET (/wishcard) route back to returning only published and donated cards.

server/routes/wishCards.js
```diff
@@ -369,11 +369,11 @@ router.get('/:id', redirectLogin, getByIdValidationRules(), validate, async (req
...
-   
-    // NOTE: 
+
+    // NOTE:
     // this agency object is returning undefined and breaking frontend
-    const agency = wishcard.wishCardTo;
-    
+    const agency = wishcard.belongsTo;
+
     console.log("test print");
```
Jacobjeevan, line 377: I think someone added this for testing (can be removed now with seeder + migrations)

gnorbsl: done

gnorbsl:
> 👍 While you are at it, do you also want to take care of the status field as well?
> 
> Ensure all existing wishcards have a status of published on prod (I think Stacey was manually setting this, but in case she missed any).
> 
> We can then revert GET (/wishcard) route back to returning only published and donated cards.

added

PR #237, opened 2020-11-29
Feature/payment success pag

Tokikko:
Implemented thank you page for a successful payment.
Route stripe/payment/success/[wishcardid]


client/views/successPayment.ejs
```diff
@@ -0,0 +1,68 @@
+<!DOCTYPE html>
+<html lang="en">
+<!-- TODO: before user can leave a message, check if user is signed in first, if not, redirect to log in -->
+
+<head>
+  <title>See Each Wish Card | Donate Gifts</title>
```
gnorbsl, line 6: wrong title?

client/views/successPayment.ejs
```diff
@@ -0,0 +1,68 @@
...
+  <!-- ** NOTE: CDN Bootstrap.js, Popper.js, and jQuery are moved to the bottom so the page loads faster-->
+  <%- include('./templates/header')-%>
+  <link rel="stylesheet" href="/public/css/successPayment.css" />
+
+  <!-- jquery -->
+  <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
```
gnorbsl, line 12: jquery is part of header I think

gnorbsl (changes requested):

server/routes/stripe.js
```diff
@@ -103,4 +104,28 @@ router.post('/webhook', bodyParser.raw({ type: 'application/json' }), async (req
   res.json({ received: true });
 });
 
+// called from frontend after stripe payment confirmation
+// redirect to a thank you page
+router.get('/payment/success/:id', redirectLogin, async (req, res) => {
+  try {
```
gnorbsl, line 110: you should make sure that only the user that donated can see this page, otherwise someone could farm all of the data by just opening this page with a wishcardid

server/routes/stripe.js
```diff
@@ -103,4 +104,28 @@ router.post('/webhook', bodyParser.raw({ type: 'application/json' }), async (req
...
+    const wishCardId  = req.params.id;
+    const wishCard = await WishCardRepository.getWishCardByObjectId(mongoSanitize.sanitize(wishCardId));
+    const totalItemPrice = await calculateWishItemTotalPrice(wishCard.wishItemPrice);
```
gnorbsl, line 113: this will change since the itemprice doesnt reflect the price if a user added money for us to the donation,

you might have to get that from the donationsmodel

gnorbsl (changes requested):

server/routes/stripe.js
Tokikko: I cant make that. Since the donation hook is inserting data into the donation and this is called immediately after payment confirmation. Plus all of the data that is visible here is visible on wishcard itself which is public

server/routes/stripe.js
Tokikko: I cant do that, same problem as above

gnorbsl (approved):

PR #249, opened 2020-12-01
Donation details page

gnorbsl:


server/routes/wishCards.js
```diff
@@ -231,20 +231,19 @@ router.get('/me', renderPermissions, async (req, res) => {
...
 // @route   GET '/wishcards/admin'
 // @access  User with admin role
 // @tested 	No
 router.get('/admin/', async (req, res) => {
   try {
-    const WISHCARD_STATUS = 'draft';
-    const USER_ROLE = 'admin';
     // only admin users can get access
-    if (res.locals.user.userRole !== USER_ROLE) {
+    if (res.locals.user.userRole !== 'admin') {
```
Tokikko, line 242: Why did you remove the constants and used pure strings?

gnorbsl: because this is a single use, constants make not much sense if you use them a single time, especially in this case where it makes everything less clear. userrole !== userrole. If you need to use constants, make them clear like ADMIN_USER_ROLE or something and then make them available file wide and not only per function.

gnorbsl: If you want to do a status enum create an object like

WISHCARD_STATUS = {
DRAFT: 'draft',
DONATED: 'donated',
}

this way you could use WISHCARD_STATUS.DRAFT this way it is very clear whats happening and what is meant. Replacing 'draft' with WISHCARD_STATUS has no advantage over just using 'draft' it just makes the code less readable.

stacysealky (approved):

server/routes/wishCards.js
Tokikko: Thats true my friend. I did not name the variables right.

PR #257, opened 2020-12-02
DON-160-LoginModal for Read more

Jacobjeevan:
- Similar logic to donate Btn
- Displays login modal for unauthenticated user
- Link for logged in user

Hmm, should we should display a similar modal + message for email unverified users (Just a message to verify their email) @gnorbsl @stacysealky  🤔


stacysealky (approved):

stacysealky:
@Jacobjeevan I don't really think we should display a modal reminder since they already see the notification in the profile page

PR #258, opened 2020-12-04
Autoprefixed css files, adjusted home, sign/login page

MarkellRichards:
Use the auto prefix command within the CSS directory. Short term fix, for now, can automate it later on with scss. 

All the CSS files show they have been changed. That's due to the auto prefix command. Adjusted home, login, sign up, & wish cards pages for cross-platform compatibility


client/public/css/home.css
```diff
@@ -130,6 +130,7 @@ New color scheme reference is in utils.css
 #home-bg {
   background: url('../img/new-hero.png') center center;
   min-height: 100vh;
+  height: 100px;
```
stacysealky, line 133: why 100px?

client/public/css/stripe.css
```diff
@@ -60,8 +60,6 @@ input {
   width: 20px;
   height: 20px;
   box-shadow: inset 0 0 0 2px;
-  -webkit-transform: translateZ(0);
```
stacysealky, line 63: why were these deleted?

client/public/css/home.css
MarkellRichards: That is a hack i found to make IE11 work when justifying content and aligning items center. It will still use the min height bUT needs a height property.

client/public/css/stripe.css
MarkellRichards: Im not sure, maybe they are no longer needed. This wasn't done manually

client/public/css/home.css
stacysealky: I had no idea! that's cool

stacysealky (approved):

PR #259, opened 2020-12-04
Added status section on homepage

deep-a:


server/app.js
```diff
@@ -209,6 +209,7 @@ const howtoRoute = require('./routes/howTo');
 const faqRoute = require('./routes/faq');
 const contactRoute = require('./routes/contact');
 const stripeRoute = require('./routes/stripe');
+const statusSection = require('./routes/homeStatusSection');
```
stacysealky, line 212: why create a different route for home here?

server/db/repository/AgencyRepository.js
```diff
@@ -17,7 +17,16 @@ async function createNewAgency(agencyParams) {
   }
 }
 
+async function getAllVerifiedAgencies() {
```
stacysealky, line 20: shouldn't this be find({isVerified: true}) ?

server/routes/homeStatusSection.js
```diff
@@ -0,0 +1,24 @@
+const express = require('express');
```
stacysealky, line 1: I don't think this should be a separate route file

stacysealky (commented):
Please fix the 2 test cases that are failing in circleCi. I think they have to do with your new route that is rendering home.ejs

server/routes/homeStatusSection.js
```diff
@@ -0,0 +1,24 @@
...
+const log = require('../helper/logger');
+
+const WishCardRepository = require('../db/repository/WishCardRepository');
+const AgencyRepository = require('../db/repository/AgencyRepository');
+
+// Renders data to homepage on sats section.
+
+router.get('/', async (req, res) => {
+  try {
+    const wCards = await WishCardRepository.getNotDonatedCards();
```
Jacobjeevan, line 13: I agree with Stacy, since this is just for one route, creating a separate file is bit overkill. Best to move the code into https://github.com/donategifts/donategifts/blob/cce033e834db222855ddd6c98c2d0762aa1dfe51/server/app.js#L187

Also, I believe this implementation is being overridden by the default home render from app.js (which does use user and wishcards).

server/db/repository/AgencyRepository.js
Jacobjeevan: This should be true, yes 🤔

server/db/repository/WishCardRepository.js
```diff
@@ -163,6 +163,14 @@ async function getWishCardByAgencyId(agencyId) {
   }
 }
 
+async function getNotDonatedCards() {
+  try {
+    return WishCard.find({ isDonated: false, status: 'published' }).exec();
```
Jacobjeevan, line 168: isDonated has been removed, so you only have to use status: 'published' 👍 

Edit: Since isDonated is removed, you can use getWishCardsByStatus instead. No need to add another method :)

Jacobjeevan (changes requested):

PR #297, opened 2020-12-24
feat: tsoa authentication setup, eslint cleanup

Enubia:


services/website-api/src/app.ts
```diff
@@ -132,11 +143,11 @@ const bootServer = async () => {
 	});
 
 	app.get(
-		'/api/website-api/openapi.yaml',
+		'/website-api/openapi.yaml',
 		cors<express.Request>(),
 		basicAuth({
 			users: {
-				docs: 'needAllInfo',
+				hurrdurrpurr: 'camelCaseWithCamelToe',
```
gnorbsl, line 150: hardcoding super secret passwords?

services/website-api/src/app.ts
```diff
@@ -155,11 +166,14 @@ const bootServer = async () => {
 
 		logger.error(err);
 
-		res.status(500);
+		res.status(500).send({
+			status: 500,
```
gnorbsl, line 170: isnt that redundant?

services/website-api/src/routes/routes.ts
```diff
@@ -340,6 +344,53 @@ export function RegisterRoutes(app: express.Router) {
 
     // WARNING: This file was auto-generated with tsoa. Please do not modify it. Re-run tsoa to re-generate this file: https://github.com/lukeautry/tsoa
 
+    function authenticateMiddleware(security: TsoaRoute.Security[] = []) {
```
gnorbsl, line 347: I know this is auto generated but shouldnt the middleware in their own file?

gnorbsl (approved):

services/website-api/src/app.ts
Enubia: that needs to be replaced with process variables, didn't do that yet

gnorbsl: now we can never use those perfect passwords again, good job.

services/website-api/src/app.ts
Enubia: nope, we only send status code on general error, this way we also send the error message

services/website-api/src/routes/routes.ts
Enubia: middleware is in its own file, just gets used in the routes since tsoa generates it that way

services/website-api/src/app.ts
gnorbsl: but couldnt you just use res.send instead of the status function

gnorbsl (approved):

PR #310, opened 2021-01-07
fix: fixed type errors and missing import

Enubia:


packages/helper/src/email/types/IEmail.ts
```diff
@@ -3,5 +3,9 @@ export interface IEmail {
 	to: string;
 	subject: string;
 	message: string;
-	attachments?: unknown;
```
Enubia, line 6: @Jacobjeevan please use 
```ts 
attachments?: any;
```
instead of unknown, unknown has side effects which cause errors
i've gone ahead and typed it for you

packages/helper/src/email/index.ts
```diff
@@ -157,7 +158,7 @@ const sendVerificationEmail = async (
 		.replace('%buttonText%', 'Confirm Your Email');
 
 	const newEmail: IEmail = {
-		from: process.env.DEFAULT_EMAIL,
+		from: process.env.DEFAULT_EMAIL!,
```
Enubia, line 161: @Jacobjeevan for process variables you can use ```!``` for non-null coearcion since those variables are always filled

packages/helper/src/email/types/IEmail.ts
Jacobjeevan: @Enubia 👍

Jacobjeevan (approved):

PR #312, opened 2021-01-08
feat: refactored user controller

Enubia:
- refactored user controller to contain all user specific stuff
- created agency controller
- added storage service for s3 file upload


packages/user/src/UserService.ts
```diff
@@ -29,32 +33,77 @@ export class UserService {
...
-		if (user) {
-			if (user.emailVerified) {
-				return false;
-			}
+		if (!user) {
+			throw new UserError('No user found for verification!');
+		}
 
+		if (!user.emailVerified) {
 			await this.userRepository.setUserEmailVerification(user._id, true);
```
Jacobjeevan, line 54: Since verifyUserEmail returns a user, shouldn't we update it on setUserEmailVerification call. Otherwise the user returned here would still have emailVerified to be  false, no? :thinking:

packages/agency/src/database/AgencyRepository.ts
```diff
@@ -1,44 +1,27 @@
...
+		return this.dbAgency.findOne({ accountManager: userId }).lean().exec();
 	}
 
 	async getAgencyWishCards(agencyId: TypeObjectId<IAgency>): Promise<IAgency | null> {
-		try {
-			return DBAgency.findOne({ _id: agencyId }).populate('wishCards').exec();
-		} catch (error) {
-			throw new Error(`Failed to get Wishcards: ${error}`);
-		}
+		return this.dbAgency.findOne({ _id: agencyId }).populate('wishCards').lean().exec();
```
Jacobjeevan, line 15: Ah, so we are going to check for it in Services, rather than using exceptions? :thinking: :+1:

Enubia: yes, i wanted to get rid of all the throws in the repositories since we catch everything usually in the service itself and the mongosetup right now logs exceptions anyways

packages/user/src/UserService.ts
Enubia: yeah that's true, that should be updated

Jacobjeevan (approved):

gnorbsl (approved):

gnorbsl:
please change the storage stuff before merging like discussed in slack

PR #316, opened 2021-01-14
refactor: moved generic express middleware helper

Enubia:
- middleware helper are now located in own package
- resolved dependency circle issues do to refactoring
- split up user in user and user-data package
- split up agency in agency and agency-data package
- fixed logger export


packages/user/src/UserService.ts
```diff
@@ -1,17 +1,17 @@
...
 import { sendPasswordResetMail } from '@donategifts/helper';
-import { UserRepository } from './database/UserRepository';
+import { UserRepository } from '@donategifts/user-data';
 import { UserError } from './helper/UserError';
 
 @injectable()
 export class UserService {
 	constructor(
-		@inject(UserRepository) private userRepository: UserRepository,
+		private userRepository: typeof UserRepository = UserRepository,
```
Jacobjeevan, line 14: Okay, qn about the whole injectable thing - My understanding is that we are injecting to get the single instance and ensure we are using that instance everywhere.....but I suppose in this case it doesn't matter since it's Repository stuff, right? :thinking: But kinda makes default scope in user data pointless, no?

Or does it automatically get the instance without using @ inject?

Enubia: it does without because the userrepository is already bound to a container
you would only need to specifically inject it if it wasn't bound to the container instance, but in the case of using it in the constructor from another package it get's bound in the origin package
Capture

Jacobjeevan (approved):

PR #319, opened 2021-01-15
Payments

Jacobjeevan:
- Added payment package for Payment Services
	- Handles webhooks for both paypal and stripe
	- Calls handleDonation to send confirmation email and message to slack
	- Types
	       - IDonationHook
	       - IStripeIntent
- Added types for few functions in wishcardRepository
	- Used in paymentHelper (and to ensure lerna works for the package and linter doesn't complain)
- Added payment controller to website api
- IDonationSlack
	- amount: changed type to number
	- userDonation: number or null
- IDonationConfirmationEmail
	- price: changed type to number
- Helper/email
	- Fixed file name error
	- Added a missing file to resources

Note: PaymentController has a ('/test') route for testing handleDonation (calls logic to send confirmation email and post to slack), which I kept,in case we change some things around. We can remove it later once things are finalized.


Jacobjeevan:
Ignore the number of line changes lol. It's almost all package-lock file.

packages/helper/src/slack/index.ts
```diff
@@ -88,6 +88,7 @@ async function sendSlackFeedbackMessage(slackEmail: ISlackFeedbackMsg): Promise<
 
 async function sendDonationNotificationToSlack(donationInfo: IDonationSlack): Promise<boolean> {
 	const { service, userDonation, donor, wishCard, amount } = donationInfo;
+	const userContribution = userDonation || 0.0;
```
gnorbsl, line 91: why 2 variable names?

packages/payment/src/StripeService.ts
```diff
@@ -0,0 +1,105 @@
...
+			const metadata: {
+				userDonation: number | null;
+				wishCardId: string | null;
+				userId: string | null;
+				agencyName: string | null;
+			} = {
+				wishCardId: wishCard._id.toString(),
+				userId: user._id.toString(),
+				agencyName,
+				userDonation: null,
```
gnorbsl, line 50: a bit pedantic but can you move userdonation to the beginning so that the order is the same in both objects

gnorbsl (changes requested):

gnorbsl:
Probably good if @Enubia takes a look at it too

packages/helper/src/slack/index.ts
Jacobjeevan: Ah, you are right, lol; not sure why I did it like that xD Will fix it :+1:

package.json
```diff
@@ -55,9 +55,11 @@
 		"eslint-import-resolver-node": "^0.3.4",
 		"eslint-import-resolver-typescript": "^2.3.0",
 		"eslint-plugin-import": "^2.22.1",
+		"eslint-plugin-jsx-a11y": "^6.4.1",
```
Enubia, line 58: the react plugins shouldn't be needed, i removed them because the latest airbnb config doesn't require them anymore

package.json
```diff
@@ -55,9 +55,11 @@
...
 		"eslint-plugin-mocha": "^8.0.0",
 		"eslint-plugin-node": "^11.1.0",
 		"eslint-plugin-prettier": "^3.3.1",
+		"eslint-plugin-react": "^7.22.0",
```
Enubia, line 62: same with this one

packages/common/src/types/payment/DonationHook.ts
```diff
@@ -0,0 +1,8 @@
+export interface IDonationHook {
+	service: string;
+	userId: string;
+	wishCardId: string;
+	amount: number;
+	userDonation: number | null;
```
Enubia, line 6: can it be ```null``` or ```undefined```? if undefined you can write it like so ```userDonation?: number;```

packages/helper/src/slack/index.ts
```diff
@@ -87,7 +87,9 @@ async function sendSlackFeedbackMessage(slackEmail: ISlackFeedbackMsg): Promise<
 }
 
 async function sendDonationNotificationToSlack(donationInfo: IDonationSlack): Promise<boolean> {
-	const { service, userDonation, donor, wishCard, amount } = donationInfo;
+	const { service, donor, wishCard, amount } = donationInfo;
+	let { userDonation } = donationInfo;
+	userDonation = userDonation || 0.0;
```
Enubia, line 92: you don't need the re assignment here, you can do the check in the template string
```/wishcards/admin/${wishCard._id}, amount: ${amount} with ${userDonation || 0.0} for us`,```

File: packages/helper/src/slack/types/IDonationSlack.ts [hunk 7]
```diff
@@ -2,8 +2,8 @@ import { IWishCard, IUser } from '@donategifts/common';
 
 export interface IDonationSlack {
 	service: string;
-	userDonation: string;
+	userDonation: number | null;
```
Enubia, line 5: same here, can it be null or undefined? if undefined you can write it like so ```userDonation?: number;```

packages/payment/src/PaymentService.ts
```diff
@@ -0,0 +1,64 @@
+import { injectable } from 'inversify';
+import * as express from 'express';
+/* import * as mongoSanitize from 'express-mongo-sanitize';
```
Enubia, line 3: remove this instead of commenting

packages/payment/src/PaymentService.ts
```diff
@@ -0,0 +1,64 @@
...
+		stripeSignature: string,
+	): Promise<{ success: boolean; error?: string }> {
+		return this.stripeService.webHook(request, stripeSignature);
+	}
+
+	public async testHandleDonation(donation: IDonationHook): Promise<boolean> {
+		return handleDonation(donation);
+	}
+
+	/* public async onSuccess(
```
Enubia, line 41: same here

packages/payment/src/PaymentService.ts
```diff
@@ -0,0 +1,64 @@
...
+import { WishCardRepository } from '@donategifts/wishcard'; */
+import { IStripeIntent /* IDonationInfo */, IDonationHook } from '@donategifts/common';
+import { StripeService } from './StripeService';
+import { PaypalService } from './PaypalService';
+import { handleDonation } from './helper/paymentHelper';
+
+@injectable()
+export class PaymentService {
+	constructor(
+		private stripeService = new StripeService(),
```
Enubia, line 14: we don't do it the JS way anymore, we use inversify to handle our injections

we inject the services with the ```@inject``` decorator to create single instances of the classes used
```ts
@inject(StripeService) private stripeService: StripService,
@inject(PaypalService) private paypalService: PaypalService
```

packages/payment/src/PaypalService.ts
```diff
@@ -0,0 +1,54 @@
...
+					const donation: IDonationHook = {
+						service: 'Paypal',
+						userId: data[0],
+						wishCardId: data[1],
+						amount: request.body.resource.purchase_units[0].amount.value,
+						userDonation: data[2],
+						agencyName: data[3],
+					};
+
+					const success: boolean = await handleDonation(donation);
```
Enubia, line 41: you don't need to exlpicitly type the succes, if handleDonation has a return type it get inherited from the function return type

packages/payment/src/PaypalService.ts
```diff
@@ -0,0 +1,54 @@
...
+					return {
+						success,
+					};
+				},
+			);
+		}
+		logger.debug('Webhook Error - Checkout order not approved');
```
Enubia, line 48: logger.error if it's an error or logger.warning

packages/payment/src/StripeService.ts
```diff
@@ -0,0 +1,105 @@
...
+		private userRepository: typeof UserRepository = UserRepository,
+	) {}
+
+	public async createIntent(
+		stripeData: IStripeIntent,
+	): Promise<{ success: boolean; error?: string; clientSecret?: string }> {
+		const { wishCardId, email, agencyName, userDonation, userId } = stripeData;
+		// Create a PaymentIntent with the order amount and currency
+		const wishCard = await this.wishCardRepository.getWishCardByObjectId(
+			mongoSanitize.sanitize(wishCardId) as TypeObjectId<IWishCard>,
```
Enubia, line 29: we shouldn't need to sanitize things here, if you still want to sanitize inputs i would do it in the controller and not the service. The service shouldn't need to sanitize typed parameters

you can also use ObjectId<IWishCard>(wishCardId) which returns the input as object id of the given type

packages/payment/src/StripeService.ts
```diff
@@ -0,0 +1,105 @@
...
+		);
+
+		const user = await this.userRepository.getUserById(
+			mongoSanitize.sanitize(userId) as TypeObjectId<IUser>,
```
Enubia, line 33: same goes here

packages/payment/src/StripeService.ts
```diff
@@ -0,0 +1,105 @@
...
+				metadata,
+			});
+
+			return {
+				success: true,
+				// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+				clientSecret: paymentIntent.client_secret!,
+			};
+		}
+		return { success: false, error: 'Wishcard not found' };
```
Enubia, line 69: I would move this right after the wishcard db call to avoid further db calls if there's not wishcard present

```ts
const wishCard = await this.wishCardRepository...

if (!wishCard) {
  return { success: false, error: 'Wishcard not found' };
}
```
this avoid hitting the db for the user if we actually don't have a wishcard

packages/payment/src/StripeService.ts
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	public async webHook(
+		request: string,
+		stripeSignature: string,
+	): Promise<{ success: boolean; error?: string }> {
+		const endpointSecret = process.env.STRIPE_SECRET || 'undefined';
```
Enubia, line 76: you can just use ```const endpointSecret = process.env.STRIPE_SECRET ;```, process variables are undefined per default if not defined anywhere and than just check for it like 
```ts
if (!endpointsecret) {
  logger.info('Webhook Error - Stripe endpoint secret not set');
  return { success: false, error: 'Webhook Error - Stripe endpoint secret not set' };
}
```

packages/payment/src/StripeService.ts
```diff
@@ -0,0 +1,105 @@
...
+						amount: event.data.object.amount / 100,
+						userDonation: metadata.userDonation,
+						agencyName: metadata.agencyName,
+					};
+					const success: boolean = await handleDonation(donation);
+					return { success };
+				}
+				logger.info('Webhook Error - Metadata not found');
+				return { success: false, error: 'Webhook Error - Metadata not found' };
+			} catch (err) {
```
Enubia, line 98: maybe log the error here as well

services/website-api/src/controller/payment/PaymentController.ts
```diff
@@ -0,0 +1,64 @@
+import * as express from 'express';
+import { PaymentService } from '@donategifts/payment';
+import {
+	/* IUser, */ IDonationHook,
```
Enubia, line 4: remove the commented stuff please

Enubia (changes requested):

packages/common/src/types/payment/DonationHook.ts
Jacobjeevan: Unfortunately, no since it's fed into stripe's API, which requires null.

packages/helper/src/slack/types/IDonationSlack.ts
Jacobjeevan: You could do it here, but I would rather stay consistent (and keep types the same, otherwise we will have to cast to undefined or number here).

services/website-api/src/controller/payment/PaymentController.ts
Jacobjeevan: It's used for the onSuccess route (which is also commented out, I didn't remove it because I wasn't sure if we were going to use it. It depends on how things are handled on frontend side).

The commented stuff on PaymentServices goes with this route :)

package.json
```diff
@@ -76,5 +75,8 @@
 		"hooks": {
 			"pre-commit": "npm run lint"
 		}
+	},
+	"dependencies": {
```
Enubia, line 79: this should be a dev dependency

packages/payment/src/StripeService.ts
```diff
@@ -0,0 +1,101 @@
...
+		// Create a PaymentIntent with the order amount and currency
+		const wishCard = await this.wishCardRepository.getWishCardByObjectId(ObjectId(wishCardId));
+
+		const user = await this.userRepository.getUserById(ObjectId(userId));
+
+		// By default stripe accepts "pennies" and we are storing in a full "dollars". 1$ == 100
+		// so we need to multiple our price by 100. Genious explanation
+		const PENNY = 100;
+		let totalItemPrice = await calculateWishItemTotalPrice(wishCard.wishItemPrice);
+		if (!wishCard) {
```
Enubia, line 35: move this up right after the wishcard query so the user doesn't get executed if there's no wishcard :D

Enubia (changes requested):
two small things than it's fine for me

Enubia:
please also merge the master since @gnorbsl s changes are for the .env loading are already merged

Jacobjeevan:
> please also merge the master since @gnorbsl s changes for the .env loading are already merged

Done :+1: Thanks for all the feedback :D

Enubia (approved):

PR #320, opened 2021-01-16
fix: load env before route registration

gnorbsl:
current setup would throw errors on start since config is loaded after route registration which clashes with the storage helper. this way would load the config from an .env file in /website-api


gnorbsl:
Also docker automatically uses the .env from the root directory of the dockerfile so you wouldnt need to define a path there either

Enubia (approved):

Issue #340, opened 2021-10-19
Add the christmas section on landing page back 


stacysealky:
currently commented out i think?

PR #344, opened 2021-10-20
added the xmas section back with new condition

stacysealky:
#340

### DIFF
--- a/client/views/home.ejs
+++ b/client/views/home.ejs
@@ -1,88 +1,84 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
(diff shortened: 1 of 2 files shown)


Jacobjeevan (approved):
Hmm, do you think we could modify the wording "DAYS UNTIL CHRISTMAS DONATIONS CLOSE" - sounds bit clunky, would days until Christmas work? Or days until holidays? Not a big deal anyways, but just my 2 cents.

Otherwise, looks good :+1:

stacysealky:
@Jacobjeevan great suggestion! I will change that and commit again. Thanks Jacob

Issue #350, opened 2021-10-20
wichcard creation form has url validation and not accepting possible formats


stacysealky:
the validation rules for url on the wishcard creation form need to be more lax
should not always include https://

PR #352, opened 2021-10-21
#350/351: Wishcard validation + Redirect

Jacobjeevan:
#351 
#350 

Changed the redirect to wishcard/me/ + Changed wishcard validation to accept http or https amazon links.

On a side note: If we have some time, I would suggest modifying wishcard/me/ a bit (add editing functionality for partners - I don't think we had the time do that last time :+1: )

There's a PUT route for updating wishcard status (from admin panel), so we can reuse that if needed.


Enubia (approved):
LGTM

PR #368, opened 2021-11-02
agency verification via slack

### LINKED ISSUE #355: Send auto-email after Agency.isVerified==true
I have a html template for the email but needs to be automated, so when db updates that agency is verified, it should automatically trigger to send the email to the accountManager

Enubia:
resolves #355


server/routes/contact.js
```diff
@@ -29,7 +29,12 @@ router.post('/email', async (req, res) => {
       message: req.body.message,
     });
 
-    const mailResponse = await sendMail(contact.email, 'stacy.sealky.lee@gmail.com', contact.subject, contact.message);
+    const mailResponse = await sendMail(
+      contact.email,
+      'stacy.sealky.lee@gmail.com',
```
stacysealky, line 34: hold on, I forgot why my email is hard coded here. why?

stacysealky (approved):

server/routes/contact.js
Enubia: I don't know, that has only changed because of eslint formatting

Issue #366, opened 2021-11-01
Move wish card creation form into a separate page


stacysealky:
Currently, agency users can submit wish cards by filling out the form within /profile page. 
Could we move this form section to a separate page? 
I think it's cleaner than having it all on the profile page.

Btw, did we decide to leave the form in /profile due to userRole security issues?

PR #369, opened 2021-11-02
extracted wish card functionality from profile into a separate page

Tokikko:
Extracted create wish card functionality from the profile into a separate createWishcard page. Was mostly just cp code from file to file :).

Create a wish card link that will be rendered only for agency users


client/views/createWishcard.ejs
```diff
@@ -0,0 +1,193 @@
...
+  <!-- CDN jQuery -->
+  <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
+
+</head>
+
+<body>
+  <%- include('./templates/nav')-%> <%- include('./templates/editAboutMeModal')-%>
+  <!-- header section-->
+ 
+  <% if (user.userRole == 'partner' && locals.agency && agency.isVerified) { %>
```
stacysealky, line 21: Does it redirect to 400 page for the ELSE case here?

server/routes/wishCards.js
```diff
@@ -232,6 +232,20 @@ router.get('/me', renderPermissions, async (req, res) => {
   }
 });
 
+// @desc    Render createWiscard.ejs. Empty form for creating a new wish card
+// @route   GET '/wishcards/new'
+// @access  Agency
+// @tested 	Yes
+router.get('/new', renderPermissions, async (_req, res) => {
```
stacysealky, line 239: Sorry for being nit picky, but can you change '/wishcards/new' to '/wiscards/create' ? 
Just because it sounds like you are viewing a new wish card instead of creating it

stacysealky (changes requested):

client/views/createWishcard.ejs
Tokikko: You are right it is not necessary here since on the server we validate if it's an agency user. Will remove it

server/routes/wishCards.js
Tokikko: yes mam

stacysealky (approved):
Closes #366

PR #513, opened 2022-12-10
added placeholder year to emails templates which will be replaced by …

Tokikko:
…current year when sending those templates


stacysealky (approved):
Thanks Tokikko!

PR #526, opened 2023-06-28
Irs proof page

stacysealky:
"501(c)(3) non profit charitable organization" text in footer has a new page link '/proof'


Enubia (approved):
LGTM!

PR #529, opened 2023-07-06
Feature/profile edit 500

stacysealky:


stacysealky:
@Enubia  - added edit features for account details and agency details, but they are not really working when I tested. Can you please review and advise me?

.gitignore
```diff
@@ -4,5 +4,5 @@
 uploads
 node_modules
 .idea
-config/config*.env
```
Enubia, line 7: please keep this, I'm playing around with a few different config options and don't want to accidentally submit a production config

design/scss/updated/footer.scss
```diff
@@ -0,0 +1,27 @@
+/*--------------------------------------------------------------
+    Footer
```
Enubia, line 2: we don't need comments like this i think, the file name speaks for itself haha

design/scss/updated/profile.scss
```diff
@@ -0,0 +1,107 @@
+/*--------------------------------------------------------------
+    Profile
```
Enubia, line 2: we don't need comments like this I think, the file name speaks for itself haha

server/controller/profile.js
```diff
@@ -91,6 +94,90 @@ module.exports = class ProfileController extends BaseController {
...
+				state,
+				country,
+				zipcode,
+			} = req.body;
+
+			if (!res.locals.user) {
+				return this.handleError(res, 403, 'No user id in request');
+			}
+
+			const accountManager = await this.#agencyRepository.getAgencyByUserId(
```
Enubia, line 141: You want to call `this.#userRepository.getUserByObjectId` right? Since you want the account manager's object to get the `_id` from him

views/pages/profile.pug
```diff
@@ -173,20 +188,89 @@ block modal
...
+                            label( for='state' ) State:
+                            input#state.mx-2.edit-input( type='text' name='state' placeholder= agency.agencyAddress.state )
+                        .row.justify-content-center.my-2
+                            label( for='country' ) Country:
+                            input#country.mx-2.edit-input( type='text' name='country' placeholder= agency.agencyAddress.country )
+                        .row.justify-content-center.my-2
+                            label( for='zipcode' ) Zipcode:
+                            input#zipcode.mx-2.edit-input( type='text' name='zipcode' placeholder= agency.agencyAddress.zipcode )
+                .modal-footer.d-flex.justify-content-center
+                    button#save-agency-details.button-primary( type='button' data-bs-dismiss='modal' ) Save
```
Enubia, line 256: please put them in the `block modal` section, the layout can be viewed in `master.pug`

views/pages/profile.pug
```diff
@@ -173,20 +188,89 @@ block modal
...
 
 block scripts
     script.
         $(document).ready(() => {
+            $('[placeholder]').focus(function() {
+                var input = $(this);
```
Enubia, line 262: you can use `let` and `const` instead of var

views/pages/profile.pug
```diff
@@ -205,6 +289,64 @@ block scripts
                 });
             });
 
+            $('#save-account-details').click(() => {
+                const fName = $('input#fName').val();
+                const lName = $('input#lName').val();
+                console.log(fName, lName)
```
Enubia, line 295: please remove this log before merging

views/pages/profile.pug
```diff
@@ -205,6 +289,64 @@ block scripts
...
+                $.ajax({
+                    type: 'PUT',
+                    url: '/profile/account',
+                    data: { fName, lName },
+                    success: (response) => {
+                        if (response.error) {
+                            return showToast('Error: Could not update account details');
+                        }
+                        showToast('Success: Updated your account details', true);
+                        console.log("response from ajax call ", response); // not updating correctly after refresh
```
Enubia, line 305: please remove this log before merging

Enubia (changes requested):

Enubia:
A few things here and there but overall good job!

Please just merge the latest master since I've updated the workflow files

server/controller/profile.js
stacysealky: I want to get the agency the user (account manager) belongs to, so that I can update their agency information.

`
const AgencySchema = new Schema(
	{
		agencyName: {
			type: String,
		},
		agencyWebsite: {
			type: String,
		},
		agencyPhone: {
			type: String,
		},
		**accountManager: {
			type: Schema.Types.ObjectId,
			ref: 'User',
		},**
		agencyBio: {
			type: String,
		},
		agencyAddress: {
			address1: { type: String },
			address2: { type: String },
			city: { type: String },
			state: { type: String },
			country: { type: String },
			zipcode: { type: String },
		},
		childrenUnderCare: {
			type: Number,
		},
		childrenAgeRange: {
			type: String,
		},
		agencyProfileImage: {
			type: String,
		},
		joined: {
			type: Date,
			default: Date.now,
		},
		isVerified: {
			type: Boolean,
			default: false,
		},
	},
	{
		collection: 'agencies',
	},
);
`

views/pages/profile.pug
stacysealky: it's already under the block modal section. line 186

stacysealky (commented):
addressed PR comments

server/controller/profile.js
Enubia: yes I noticed that, but, you actually don't need to get the user either since you already have the user._id in the response object `res.locals.user._id`

and the `updateAgency` function takes as first parameter the user id

views/pages/profile.pug
Enubia: oh right sorry, didn't see that at first glance

server/controller/profile.js
stacysealky: Ok fixed the param issue

Enubia (approved):

PR #536, opened 2023-07-09
wish card single page redesigned

stacysealky:


Enubia (changes requested):
LGTM! Why did you delete the seeder-data files tho? They are needed to fill the local db and I'd like to use them for the postgres setup as well

stacysealky:
> LGTM! Why did you delete the seeder-data files tho? They are needed to fill the local db and I'd like to use them for the postgres setup as well

I only deleted the JSON files auto generated from importing. They were supposed to be removed, no? You had a line that says "rm -r *.json" or something inside mongoimport.sh when I saw that I removed them. 
They are not anything that you previously placed in the seeder-data.

Enubia:
> > LGTM! Why did you delete the seeder-data files tho? They are needed to fill the local db and I'd like to use them for the postgres setup as well
> 
> I only deleted the JSON files auto generated from importing. They were supposed to be removed, no? You had a line that says "rm -r *.json" or something inside mongoimport.sh when I saw that I removed them. They are not anything that you previously placed in the seeder-data.

Ah yeah sorry, I forgot to remove that one line from the script. They are supposed to stay in the repo now since they can be used to fill the local database

PR #542, opened 2023-07-12
fix: fixed modals

Enubia:
added stylelint stuff and fixed the bootstrap compilation


design/scss/legacy/utils.scss
```diff
@@ -114,16 +114,16 @@ Updated buttons
...
 	border: none;
-	font-weight: $semi-bold;
+	font-weight: $fw-semi-bold;
 }
 
 .button-accent {
 	@extend %btn;
 	@extend %btn-updated;
-	background-color: $custom-orange;
+	background-color: $secondary;
```
stacysealky, line 123: Can you change the variable name from $secondary to $accent? 
I'm trying to have 3 buttons (button-accent for any orange bg, button-primary for navy bg, button-secondary for unimportant things like "log out" button on profile page, which is gray bg with navy color)

design/scss/style.scss
```diff
@@ -45,71 +45,76 @@ move this later -- from home.ejs
...
 	.input-block {
 		display: block;
 		width: 100%;
 		background: $white;
 		line-height: 1.75;
 		color: $black;
 		padding: 0.5rem 1.25rem;
-		border: 0.1rem solid $custom-orange;
+		border: 0.1rem solid $secondary;
 		border-radius: 0.5rem;
```
stacysealky, line 82: $accent = $custom-orange
$secondary = $navy

design/scss/updated/profile.scss
```diff
@@ -1,70 +1,78 @@
...
 		background-size: cover;
 		min-height: 33vh;
 		display: flex;
 		justify-content: center;
 		align-items: center;
+	
 		h1 {
-			@include font-styling($font-heading, 3rem, $custom-orange);
+			@include font-styling($font-heading, 3rem, $secondary);
+
```
stacysealky, line 16: here too

Enubia:
@stacysealky I've added a theme page where we can check all the components.
> Can you change the variable name from $secondary to $accent? I'm trying to have 3 buttons (button-accent for any orange bg, button-primary for navy bg, button-secondary for unimportant things like "log out" button on profile page, which is gray bg with navy color)

I want to go with the default bootstrap conventions if you don't mind since this is easier to use for other people once we get some help.

You can check it in the theme page. Checkout this branch and run `npm run dev` then navigate to `/theme`

stacysealky:
> @stacysealky I've added a theme page where we can check all the components.
> 
> > Can you change the variable name from $secondary to $accent? I'm trying to have 3 buttons (button-accent for any orange bg, button-primary for navy bg, button-secondary for unimportant things like "log out" button on profile page, which is gray bg with navy color)
> 
> I want to go with the default bootstrap conventions if you don't mind since this is easier to use for other people once we get some help.
> 
> You can check it on the theme page. Check out this branch and run `npm run dev` then navigate to `/theme`
> 
> As you will see, a lot of old CSS breaks some of the components, so this can be used as a reference as well to fix legacy stuff.

Ok, but 
$accent = orange bg -> used for highlighted action, and other special attentions
$primary = navy bg -> used for donate gift buttons and most primary action buttons
$secondary = gray bg -> used for secondary support actions, and actions that are not important

so it's still following the bootstrap convention. Bootstrap supports $accent color.

Enubia:
Ah I see I will change that of course

PR #547, opened 2023-07-14
531-update-stripe-npm-and-client-side-js-to-latest

Enubia:
update stripe packages
added types for stripe and paypal
updated paymentsuccess page, it does not use custom css anymore and looks the same


Enubia:
Screen Shot 2023-07-14 at 12 04 41
Screen Shot 2023-07-14 at 12 06 35

PR #581, opened 2023-08-01
refactor: api-controller

### LINKED ISSUE #556: TASK: move api specific code to own controller
some of the API code is still contained in the old controllers

create a new API controller for every API route that is being used right now and move the code

Enubia:
resolves #556

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,9 +2,11 @@
 . "$(dirname -- "$0")/_/husky.sh"
 
 export NVM_DIR="$HOME/.nvm"
-[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
 
-nvm use
+if [ -f .nvmrc ] && [ -s "$NVM_DIR/nvm.sh" ];
+then
+    . "$NVM_DIR/nvm.sh"
+    nvm use
+fi;
 
-# husky pre commit hooks are not working correc
 npx lint-staged
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,5 +1,10 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
+export NVM_DIR="$HOME/.nvm"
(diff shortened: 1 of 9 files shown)


.husky/pre-commit
```diff
@@ -1,5 +1,10 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
+export NVM_DIR="$HOME/.nvm"
```
Enubia, line 4: not to sure if this would work on machines without nvm installed, maybe someone has a better idea

Enubia: No, the fix I added does not require it to be present anymore

It was previously but now if it can't find nvm it just executes npx lint-staged without changing the node version first

nick-w-nick: Gotcha, that's what I assumed but just wanted to make sure. Thanks!

src/api/controller/basecontroller.ts
```diff
@@ -0,0 +1,48 @@
...
+
+		if (renderErrorPage) {
+			res.status(statusCode).render(code === 400 ? '404' : code.toString(), {
+				statusCode,
+				error,
+			});
+		} else {
+			res.status(statusCode).send({
+				error,
+			});
```
nick-w-nick, line 45:
```suggestion
			return res.status(statusCode).render(code === 400 ? '404' : code.toString(), {
				statusCode,
				error,
			});
		} else {
			return res.status(statusCode).send({
				error,
			});
```

Any specific reason not to return the response here? Technically the `.send` method does terminate the request lifecycle already, but this ensures the function manually exits instead of needing to check the call stack again.

Enubia: Actually good point. I've never thought about that before since I'm used to just not return when the last statement terminates.

nick-w-nick: It's a convention I've grown fond of using over time to indicate that the operation being returned is in fact the last operation that needs to be processed.

Without a hard return, it makes it seem like some additional logging/metrics will be captured once the error response is returned, but in this case, it is not.

I'd be happy to make a PR to add it to any of the necessary places if this pattern works for you.

Issue #575, opened 2023-07-29
DEFECT: /profile page edit features not updating properly


stacysealky:
Fix /profile page edit features

- [ ] when user clicks the pencil icons in the /profile page, edit modals should populate, then after saving the data, it should update immediately.

stacysealky:
Assigned to @nguyenit67

nguyenit67:
do you assign deadline to us or we assign it ourselves

Enubia:
@nguyenit67 no deadline specified, so take your time

nguyenit67:
resolved frontend bug in #583

nguyenit67:
@stacysealky the bug had been resolved in PR #585

stacysealky:
@nguyenit67 - Can you please test this in dev server? https://donate-gifts-test.fly.dev/

nguyenit67:
Kay, I will check it once I got home

stacysealky:
@nguyenit67 I closed this one and moved it to 'done'.  I added a new issue #603

PR #583, opened 2023-08-02
fix frontend bug for /profile page edit features

nguyenit67:
I've fix the frontend bug for /profile page edit features, but when I reload the page, the user account infomation (fName, lName, aboutMe) didn't update although the data has been sent in the request payload when I opened DevTools (I also check if the info has been updated in Mongo Compass, yes it was), and only when I sign out and log in again the information becomes up-to-date

### DIFF
--- a/views/pages/profile/overview.pug
+++ b/views/pages/profile/overview.pug
@@ -6,8 +6,7 @@ block title
 block content
     .profile
         .profile-welcome.cool-font.center-elements.text-secondary
-            - var title = 'Welcome' + ' ' + user.fName;
-            h1= title
+            h1#welcome-title Welcome #{user.fName}
 
         .container
             if !user.emailVerified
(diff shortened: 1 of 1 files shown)


nguyenit67:
I suspect the frontend code use the old user data in the browser storage (cookie, localStorage) and the problem lies in this line https://github.com/donategifts/donategifts/blob/master/src/index.ts#L108

Enubia (approved):
LGTM!

Enubia:
> I suspect the frontend code use the old user data in the browser storage (cookie, localStorage) and the problem lies in this line https://github.com/donategifts/donategifts/blob/master/src/index.ts#L108

Yes this is correct, the session data isn't updated correctly. Please raise an issue to fix this behavior.

nguyenit67:
I've raised the issue at #584, please check it out

stacysealky:
resolves #575

Issue #584, opened 2023-08-03
DEFECT: user account data not update properly in local session (browser/server) although updated in MongoDB


nguyenit67:
when reload the page, the user account infomation (fName, lName, aboutMe) didn't update although the data has been sent in the request payload in DevTools (the account data has also been updated in MongoDB, checked in MongoDB Compass), and just only when log out and log in again the information becomes up-to-date

Mentioned in: pull request #583 for fixing issue #575

Suspect: the problem lies in this line of code 
https://github.com/donategifts/donategifts/blob/master/src/index.ts#L108

PR #585, opened 2023-08-03
Fix session user data not update properly after calling /profile update routes and updated in MongoDB

nguyenit67:
solves #584  and #575

### DIFF
--- a/src/db/repository/UserRepository.ts
+++ b/src/db/repository/UserRepository.ts
@@ -25,11 +25,10 @@ export default class UserRepository {
 
 	async getUserAndUpdateById(id: string, updateParams: Partial<User>) {
 		try {
-			const updatedUser = await this.userModel
+			return await this.userModel
 				.findOneAndUpdate({ _id: id }, { $set: updateParams }, { new: true })
 				.lean()
 				.exec();
-			return updatedUser;
 		} catch (error) {
 			throw new Error(`Failed to update user: ${error}`);
 		}
--- a/src/db/repository/UserRepository.ts
+++ b/src/db/repository/UserRepository.ts
@@ -23,6 +23,18 @@ export default class UserRepository {
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


src/db/repository/UserRepository.ts
```diff
@@ -23,6 +23,18 @@ export default class UserRepository {
 		}
 	}
 
+	async getUserAndUpdateById(id: string, updateParams: Partial<User>) {
+		try {
+			const updatedUser = await this.userModel
+				.findOneAndUpdate({ _id: id }, { $set: updateParams }, { new: true })
+				.lean()
+				.exec();
+			return updatedUser;
```
Enubia, line 32: You can just return the update statement, no need to assign the result first

Enubia (changes requested):

src/db/repository/UserRepository.ts
nick-w-nick, line 32:
```suggestion
			return await this.userModel
				.findOneAndUpdate({ _id: id }, { $set: updateParams }, { new: true })
				.lean()
				.exec();
```

@nguyenit67 here's a suggestion you can just apply to match @Enubia's change request so you don't have to do it manually.

Enubia: Thanks for the change suggestion, couldn't figure out how to do that in the mobile app haha

nguyenit67: Ok I think that will make the code more concise and aligned with existed codebase

nguyenit67: @nick-w-nick thanks pal!

nguyenit67: Also, what's your opinion about assigning return value to an easy to understand name variable, for further debug/logging/doing more stuff using that variable? I want to learn more about coding convention/style guide from an experienced dev like you, thanks pretty much

Enubia (approved):

PR #586, opened 2023-08-03
Add explicit return for response handlers

nick-w-nick:
Per [this](https://github.com/donategifts/donategifts/pull/581#discussion_r1282087648) discussion on PR #581, we would like to apply a blanket change to add explicit returns to all response handlers to ensure the call stack terminates without relying on the library functions to handle the exit themselves.

I've tested a majority of these routes and did not experience any unexpected or invalid behavior. We should also look into a linting rule that makes this a requirement if possible.

### DIFF
--- a/src/api/controller/agency.ts
+++ b/src/api/controller/agency.ts
@@ -22,7 +22,7 @@ export default class AgencyController extends BaseController {
 				return this.handleError(res, 'Agency could not be found');
 			}
 
-			this.sendResponse(res, agency);
+			return this.sendResponse(res, agency);
 		} catch (error) {
 			return this.handleError(res, error);
 		}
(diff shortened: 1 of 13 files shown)


Enubia (approved):
LGTM!

You can add the eslint rule if you want, otherwise we'll create a ticket for it

stacysealky:
@nick-w-nick could you resolve the merge conflicts?

nick-w-nick:
@stacysealky All merge conflicts have been resolved, thanks for the heads up.

@Enubia I've added the rule to enforce explicit returns via TS rather than ESLint to prevent ESLint from arbitrarily requiring returns since it doesn't have introspective into the type system.

This did result in a handful more changes for some of the places I missed initially as I was only targeting direct API response statements. Feel free to check over them and let me know if you notice any issues prior to merging.

PR #604, opened 2023-08-08
fix buttons in mobile view

### LINKED ISSUE #601: DEFECT: fix home buttons
recent additions from #537 broke the home buttons as can be seen here

image

please fix them and check what else is broken due to those changes

stacysealky:
added mobile responsiveness 

resolves #601


views/pages/home.pug
```diff
@@ -23,10 +23,11 @@ block content
...
                         span.typing( data-wait='2000' data-words='["Foster Children", "Homeless Youth", "Kids in Emergency Shelters"]' )
-                    p.py-1.mb-0 Every child should feel valued and loved
-                    p.py-1.mb-5 Donate holiday gifts and school supplies to the kids in need
-                    a.button-outline.m-3( href='/howto' ) How It Works
-                    a.button-accent.m-3( href='/wishcards' ) See Wish Cards
+                    p.py-1.mb-0.mobile-hide Every child should feel valued and loved
+                    p.py-1.mb-4.mobile-hide Donate holiday gifts and school supplies to the kids in need
+                    .d-flex.justify-content-center.align-items-center.actions
+                        a.button-outline.m-2( href='/howto' ) How It Works
+                        a.button-accent.m-2( href='/wishcards' ) See Wish Cards
```
Enubia, line 30:
```suggestion
                    p.py-1.mb-0.d-none.d-md-block Every child should feel valued and loved
                    p.py-1.mb-4.d-none.d-md-block Donate holiday gifts and school supplies to the kids in need
                    .d-md-flex.justify-content-center.align-items-center
                        a.btn.btn-outline-light.btn-lg.m-2.w-75( href='/howto' ) How It Works
                        a.btn.btn-secondary.btn-lg.m-2.w-75( href='/wishcards' ) See Wish Cards
```

Enubia: achieves the same and uses bootstrap stuff

image

design/scss/custom/_home.scss
```diff
@@ -104,3 +108,13 @@
...
+
+@media (width <= 768px) {
+	.actions {
+		flex-flow: column;
+	}
+
+	.mobile-hide {
+		display: none;
+	}
+}
```
Enubia, line 120: please use bootstrap classes instead of custom breakpoints

https://getbootstrap.com/docs/5.3/utilities/flex/
https://getbootstrap.com/docs/5.3/utilities/display/

views/pages/home.pug
Enubia: image

Enubia (changes requested):

Enubia:
@stacysealky if you don't want to use the bootstrap buttons because of naming conventions we can of course change that

but I'd recommend you use them since they are fully responsive and don't need any additional styling attached to them just to make the mobile view work

views/pages/home.pug
stacysealky: but it's not the exact look I'm going for though. the buttons are not exact match

stacysealky:
> @stacysealky If you don't want to use the bootstrap buttons because of naming conventions we can of course change that
> 
> I'd recommend using them since they are fully responsive and don't need any additional styling attached to them just to make the mobile view work
> 
> We can also change the look of them a bit (more spacing and stuff like that)

yes I really don't like the naming.
also the look is not the exact match of what I designed when I use bootstrap.
will change everything except the custom buttons.

stacysealky:
@Enubia  I will not change to bootstrap buttons because they do not look the same as the design I want specifically. 
I removed the .mobile-hide class because that one is a valid point.

Enubia (approved):

PR #606, opened 2023-08-09
Reusable empty modal

NMilivojevic:
Initial modal is created, it accepts modalTitle, modalFooter, modalBody and modalRIghtSide (the orange part of signup/login modals), this is the first iteration that works only for Login component which is made but not rendered due to inability to control its open/close state with current pug nav (I left the comment in js->react->bundle.jsx), but it will be possible as soon as pug nav is converted into react component. We will modify modal component accordingly as it gets more and more complex. Let me know if something seems off.


js/react/components/shared/Modal.jsx
```diff
@@ -0,0 +1,50 @@
+import PropTypes from 'prop-types';
+
+function Modal({ modalTitle, modalBody, modalFooter, rightSide }) {
```
Enubia, line 3: since this is the modal you can name the props without the modal prefix

```suggestion
function Modal({ title, body, footer, sideContent }) {
```

if it's named sideContent we can extend the functionality in the future and place it to either right or left side

Enubia (changes requested):

PR #607, opened 2023-08-09
Convert nav into react component

NMilivojevic:
Should test the conditionals regarding if the user is logged in, is admin, is partner, etc, to see if I have successfully converted that part. Added #header-root in header part of master pug file


js/react/components/nav/index.jsx
```diff
@@ -0,0 +1,154 @@
...
+										Admin Panel
+									</a>
+								</li>
+							) : (
+								<li className="nav-item">
+									<a className="nav-link" href="/profile">
+										Profile
+									</a>
+								</li>
+							)}
```
Enubia, line 68: is it possible to do this condition without nested ternary expressions?
it's really hard to read and understand in this way

js/react/components/nav/index.jsx
```diff
@@ -0,0 +1,154 @@
+import PropTypes from 'prop-types';
+
+function Nav(props) {
```
Enubia, line 3: please try to create a new navbar with this in mind 
https://getbootstrap.com/docs/5.3/components/navbar/#responsive-behaviors

right now we have two separate ones but bootstrap is able to use only one navbar and collapse it on mobile

Enubia (changes requested):

example-config.env
```diff
@@ -1,44 +0,0 @@
-PORT=3000
-
```
Enubia, line 2: why was this file deleted?

js/react/components/nav/index.jsx
NMilivojevic: Maybe with explicit && conditionals. 
Screenshot from 2023-08-09 14-39-22

example-config.env
NMilivojevic: Forgot to copy it and then rename it, my bad

Enubia (approved):
LGTM!

Issue #605, opened 2023-08-09
TASK: validate wishItemURL from wishcard


stacysealky:
Page https://donate-gifts.com/wishcards/create

We will get the product details from the wishItemURL of WishCard Schema.

Example URL 1: https://www.amazon.com/dp/B09FPHSK74/ref=cm_gf_acsu_d_p0_e0_qd0_uaUeabH0QpVlPXTxWyIT?th=1
Example URL 2: https://www.amazon.com/Boxgear-Inseparable-Creative-Preschool-Activities/dp/B0CC31TKY2/ref=sr_1_5?keywords=toy&qid=1691540699&sr=8-5

Product ID: B09FPHSK74, B0CC31TKY2

String inside followed by "dp/" is the product ID. 

- [x] Task 1: validate string not empty
- [x] Task 2: validate string starts with "https://www.amazon.com/"
- [x] Task 3: validate that the product ID exists in the string - if productID is not found, reject submission!
- [x] Task 4: extract and save the productId separately when posting the wish card data object ---> first need to add "productID" field into the db/models/WishCard.ts

Tokikko:
Il take this... (maybe)

stacysealky:
> Il take this... (maybe)

Awesome, assigned!

PR #608, opened 2023-08-09
Update validation for wishcard to check for product id in url, save product …

Tokikko:
…id to with wishcard db


stacysealky (approved):
Looks good! resolves #605 
Could you change the text?

src/middleware/validations.ts
```diff
@@ -196,12 +197,9 @@ export default class Validations {
...
-			body('wishItemURL')
-				.notEmpty()
-				.withMessage('Wish item url is required')
-				.isString()
-				// https://regex101.com/r/yM5lU0/13 if you want to see it in action
-				.matches(/^(https?(:\/\/)){1}([w]{3})(\.amazon\.com){1}\/.*$/)
+			amazonValidator
+				.body('wishItemURL')
+				.isValidLink()
 				.withMessage('Wish item url has to be a valid amazon link!'),
```
stacysealky, line 203: Could you change this to "Item URL must start with https://www.amazon.com/ to be a valid Amazon link."

I think more information helps the user to fix their error.

stacysealky (changes requested):

src/middleware/validations.ts
Tokikko: Done :)

src/validator/amazonLink.validator.ts
```diff
@@ -0,0 +1,18 @@
+import { ExpressValidator } from 'express-validator';
```
Enubia, line 1: move into validations class please

Enubia (changes requested):

src/validator/amazonLink.validator.ts
Tokikko: Maybe extract validations for each route in its own file then to keep everything in one place?

Enubia: we can do that too, sure

I'd suggest putting them into a folder (validations or something like that) inside middleware and naming the files according to the route

It probably needs some refactoring to have a base class for error handling and the validate function since this is used in all validation classes and extend all other validation classes from this one

Tokikko: Okay, il move this for now and we should create a separate ticket for  this.

Enubia (approved):

PR #609, opened 2023-08-10
Fix /profile fields update

### LINKED ISSUE #603: DEFECT: profile edit leftover tasks
- [ ] profile edit for agency description is not updating immediately after save
- [ ] validations for empty strings missing for some fields in the edit modal -> please test and add validation rules

nguyenit67:
Fix profile fields not update immediately after save 
and add validation rules for updating account and agency fields

resolves #603


Enubia (approved):
LGTM!

@stacysealky please verify the phone regex as I'm not familiar with US phone numbers, merge if it's fine for you

src/middleware/validations.ts
```diff
@@ -107,27 +107,58 @@ export default class Validations {
...
-			body('fName', 'First name is required!').exists(),
-			body('lName', 'Last name is required!').exists(),
+			body('fName', 'First name is required!').trim().isLength({ min: 1 }).isString(),
+			body('lName', 'Last name is required!').trim().isLength({ min: 1 }).isString(),
 		];
 	}
 
+	private static readonly AGENCY_PHONE_NUMBER_REGEX =
+		/^\(?([0-9]{3})\)?([ .-]?)([0-9]{3})\2([0-9]{4})/;
+
```
stacysealky, line 122: could you change it to 
^[(]?[2-9][0-9]{2}[)]?[-\s]?[0-9]{3}[-\s]?[0-9]{4}$

US phone numbers cannot start with 0 or 1, by the way.

stacysealky (changes requested):
Great work! Just one change requested for the regex :) Thanks

src/middleware/validations.ts
```diff
@@ -107,27 +107,58 @@ export default class Validations {
...
 		return [
-			body('fName', 'First name is required!').exists(),
-			body('lName', 'Last name is required!').exists(),
+			body('fName', 'First name is required!').trim().isLength({ min: 1 }).isString(),
+			body('lName', 'Last name is required!').trim().isLength({ min: 1 }).isString(),
 		];
 	}
 
+	private static readonly AGENCY_PHONE_NUMBER_REGEX =
+		/^\(?([0-9]{3})\)?([ .-]?)([0-9]{3})\2([0-9]{4})/;
```
nguyenit67, line 121: ofc, this is the new code change, please take a look and check if it's correct yet
```suggestion
	private static readonly AGENCY_PHONE_NUMBER_REGEX =
		/^[(]?[2-9][0-9]{2}[)]?[-\s]?[0-9]{3}[-\s]?[0-9]{4}$/;
```

src/middleware/validations.ts
nguyenit67: of course, this is the new code change, please take a look and check if it's correct yet
```suggestion
	private static readonly AGENCY_PHONE_NUMBER_REGEX =
		/^[(]?[2-9][0-9]{2}[)]?[-\s]?[0-9]{3}[-\s]?[0-9]{4}$/;

```

nguyenit67, line 122: @stacysealky of course, this is the new code change, please take a look and check if it's correct yet
```suggestion
	private static readonly AGENCY_PHONE_NUMBER_REGEX =
		/^[(]?[2-9][0-9]{2}[)]?[-\s]?[0-9]{3}[-\s]?[0-9]{4}$/;

```

Enubia: I actually would update it a bit further

```js
/^(\([2-9][0-9]{2}\)[-\s]|[2-9][0-9]{2}[-\s])?[0-9]{3}[-\s]?[0-9]{4}$/
```

this matches the following strings

`123-1222`
`(234)-123-1222`
`234-123-1222`
`(234) 123-1222`
`234 123-1222`

your solution would also match `(223-123-1222` `223)-123-1222` which is not correct I think

src/middleware/validations.ts
Enubia, line 121:
```suggestion
		/^(\([2-9][0-9]{2}\)[-\s]|[2-9][0-9]{2}[-\s])?[0-9]{3}[-\s]?[0-9]{4}$/;
```

Enubia (changes requested):

src/middleware/validations.ts
Enubia: @stacysealky @nguyenit67

I actually would update it a bit further

```js
/^(\([2-9][0-9]{2}\)[-\s]|[2-9][0-9]{2}[-\s])?[0-9]{3}[-\s]?[0-9]{4}$/
```

this matches the following strings

`123-1222`
`(234)-123-1222`
`234-123-1222`
`(234) 123-1222`
`234 123-1222`

your solution would also match `(223-123-1222` `223)-123-1222` which is not correct I think

stacysealky: Hmm but mine doesn't match (223-123-1222 223)-123-1222. Did you test it?

Enubia: it does

yours matches this `(223-123-1222` and this `223)-123-1222`

stacysealky: Yeah ok, let's just get rid of the brackets and unify the format to 333-333-3333 only
That way we can add a mask to the input UI

^[2-9][0-9]{2}[-][0-9]{3}[-][0-9]{4}$

stacysealky: Also I thought you meant that it matches "(223-123-1222 223)-123-1222"
Didn't realize they were two separate strings @Enubia

stacysealky: @Enubia @nguyenit67  Let's just go with what Marco wrote
Sorry for the confusion haha
/^(\([2-9][0-9]{2}\)[-\s]|[2-9][0-9]{2}[-\s])?[0-9]{3}[-\s]?[0-9]{4}$/

Enubia:
@stacysealky updated the regex, not sure if @nguyenit67 is available right now

nguyenit67:
Sr for the late reply, I just finish my last subject exam

nguyenit67:
Sr for the late reply, I've just finished my last subject's exam. Do we need to support the international phone number format also? @Enubia @stacysealky

PR #620, opened 2023-08-13
feat: added kysely

### LINKED ISSUE #615: TASK: Integrate Kysely https://github.com/kysely-org/kysely

Enubia:
- added kysely
- created connections stuff
- updated init.sql to reflect latest changes (uuid for primary keys and constraint fixes)
- added introspection script to generated latest db types

resolves #615


types/database.ts
```diff
@@ -0,0 +1,165 @@
...
+	created_at: Generated<Timestamp>;
+}
+
+export interface Items {
+	id: Generated<string>;
+	name: string;
+	price: Numeric;
+	link: string;
+	retailer: string;
+	product_id: string;
```
stacysealky, line 75: @Enubia 
let's add color and size string fields too 
I thought about the variation details but for most cases, we just need color and size details for the order
no need to complicate it by making a meta_data json field

Enubia: > Let's add color and size string fields too

If we want to add something like this we should go with the `meta_data` field.
This way we are future-proof in case we want to add additional fields and don't need to write migrations for those cases.

stacysealky: but why? it will be harder to separate string inputs, as you don't know what the type is supposed to be.
we do not need to accommodate anything more than the color and size
and it will stay the same in the future too
@nick-w-nick do you have an insight on this?

stacysealky:
> @stacysealky @nick-w-nick updated items table with `meta_data` JSON field I've also renamed the `product_id` column to `retailer_product_id` so it's distinct from our internal naming convention. This way it cannot be confused with a constraint to another table

ok but can you please provide your reason to why we need more than the size and color variation details? meta_json seems like we would have more difficulty filtering the input data since we don't know what will be in there

Enubia:
> > @stacysealky @nick-w-nick updated items table with `meta_data` JSON field I've also renamed the `product_id` column to `retailer_product_id` so it's distinct from our internal naming convention. This way it cannot be confused with a constraint to another table
> 
> ok but can you please provide your reason to why we need more than the size and color variation details? meta_json seems like we would have more difficulty filtering the input data since we don't know what will be in there

I just want to be future-proof and avoid having to write migrations honestly.
Checking the JSON isn't that bad since we will know what keys are saved and can just iterate through them.
If we keep this project growing with additional services then this table will blow up in size at some point in the future.

types/database.ts
nick-w-nick: @stacysealky The only issue is that once we add other retailers they may not have the same variation options. Color and size are definitely a majority of them, but I could see multiple other types being relevant in the future, such as style, pattern, version, etc. The reason a simple metadata field that allows for free-form input is more fitting for the time being is:
- We don't restrict the agency to using a specific set of field for variation details in case it doesn't fit the color/size type
- We are simply adding this field as a trial run for agencies, I believe we decided that even the metadata field wasn't the most optimal solution, but by adding it we gain an opportunity for gathering data on how the agencies use it and based on that data we can decide if we need to add standardized variation fields like color/size as you mentioned, or if we need to find a different way to allow them to specify the details

Remember, once we integrate with Amazon, it will automatically populate the item on the page once they add the link, and will show the selected variation before they create the wishcard so they can verify they selected the correct one before proceeding.

Basically, we don't want to add static fields that relate to specific variations, as we aren't even sure how/if the agency plans to supply those details during creation and it will be difficult to move from specific fields to a generic field rather than vice versa if we decide to do so.

scripts/init.sql
```diff
@@ -1,158 +1,146 @@
...
+    phone varchar(100) NOT NULL,
+    bio varchar(500),
+    address_line_1 varchar(255) NOT NULL,
+    address_line_2 varchar(255),
+    city varchar(100) NOT NULL,
+    state varchar(100) NOT NULL,
+    country varchar(100) NOT NULL,
+    zip_code varchar(50) NOT NULL,
+    verified boolean NOT NULL DEFAULT false,
+    employer_identification_number varchar(255) NOT NULL,
```
nick-w-nick, line 16: How will this work for as `NOT NULL` for existing agencies in the DB that we don't yet have EIN's for?

scripts/init.sql
```diff
@@ -1,158 +1,146 @@
...
-        created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
-        updated_at timestamp,
-        CONSTRAINT wishcards_pkey PRIMARY KEY(id)
-    );
-
--- constraints
+SET check_function_bodies = false ;
+
+CREATE TABLE agencies(
+    id uuid NOT NULL DEFAULT gen_random_uuid(),
```
nick-w-nick, line 4:
```suggestion
    agency_id uuid NOT NULL DEFAULT gen_random_uuid(),
```

@Enubia What are your thoughts on having table-specific ID columns like this so we aren't consistently dealing with the ambiguous column names error when doing joins?

Enubia: I'm used to the primary key being called `id`.
Personally, I like `<variable>.id` more.
If I, for example, need to access the user id, it would be a bit redundant to call `user.user_id` in my opinion.
But let's see what @stacysealky thinks. I'm open to changes of course.

scripts/init.sql
```diff
@@ -1,158 +1,146 @@
...
+    state varchar(100) NOT NULL,
+    country varchar(100) NOT NULL,
+    zip_code varchar(50) NOT NULL,
+    verified boolean NOT NULL DEFAULT false,
+    employer_identification_number varchar(255) NOT NULL,
+    account_manager_id uuid NOT NULL,
+    image_id uuid NOT NULL,
+    created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    updated_at timestamp,
+    CONSTRAINT agencies_pkey PRIMARY KEY(id)
```
nick-w-nick, line 21:
```suggestion
    CONSTRAINT agencies_pkey PRIMARY KEY(id),
    UNIQUE(employer_identification_number)
```

We should add a unique constraint for the EIN number just for the sake of compliance.

scripts/init.sql
nick-w-nick: @Enubia It is agency-based, yes, each individual agency will have its own EIN. Individual users inside of an agency will not.

We should go through the process of getting the EIN for all existing agencies and once we can fill in the gaps we then add the `NOT NULL` constraint after the fact. We will just need to make sure we add validation on the backend that prevents null entries in the meantime.

scripts/init.sql
```diff
@@ -1,158 +1,147 @@
...
+    bio varchar(500),
+    address_line_1 varchar(255) NOT NULL,
+    address_line_2 varchar(255),
+    city varchar(100) NOT NULL,
+    state varchar(100) NOT NULL,
+    country varchar(100) NOT NULL,
+    zip_code varchar(50) NOT NULL,
+    verified boolean NOT NULL DEFAULT false,
+    employer_identification_number varchar(255) NOT NULL,
+    account_manager_id uuid NOT NULL,
```
nick-w-nick, line 17:
```suggestion
    account_manager_id uuid NOT NULL REFERENCES users (id),
```
added account manager foreign key constraint

scripts/init.sql
```diff
@@ -1,158 +1,147 @@
...
+    image_id uuid NOT NULL,
```
nick-w-nick, line 18:
```suggestion
    image_id uuid NOT NULL REFERENCES images (id),
```

added images foreign key constraint

scripts/init.sql
```diff
@@ -1,158 +1,147 @@
...
+    story varchar(255) NOT NULL,
+    image_id uuid NOT NULL,
+    created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    CONSTRAINT children_pkey PRIMARY KEY(id)
+);
+
+CREATE TABLE community_posts(
+    id uuid NOT NULL DEFAULT gen_random_uuid(),
+    message varchar(255) NOT NULL,
+    agency_id uuid NOT NULL,
```
nick-w-nick, line 40:
```suggestion
    agency_id uuid NOT NULL REFERENCES agencies (id),
```

added agency foreign key constraint

scripts/init.sql
```diff
@@ -1,158 +1,147 @@
...
+    image_id uuid NOT NULL,
```
nick-w-nick, line 41:
```suggestion
    image_id uuid NOT NULL REFERENCES images (id),
```

added image foreign key constraint

scripts/init.sql
```diff
@@ -1,158 +1,147 @@
...
+
+CREATE TABLE items(
+    id uuid NOT NULL DEFAULT gen_random_uuid(),
+    "name" varchar(255) NOT NULL,
+    price numeric(10, 2) NOT NULL,
+    link varchar(2000) NOT NULL,
+    retailer varchar(255) NOT NULL,
+    retailer_product_id varchar(255) NOT NULL,
+    meta_data json,
+    image_id uuid NOT NULL,
```
nick-w-nick, line 62:
```suggestion
    image_id uuid NOT NULL REFERENCES images (id),
```

added image foreign key constraint

scripts/init.sql
```diff
@@ -1,158 +1,147 @@
...
+    created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    CONSTRAINT items_pkey PRIMARY KEY(id)
+);
+
+CREATE TABLE messages(
+    id uuid NOT NULL DEFAULT gen_random_uuid(),
+    "content" varchar(500) NOT NULL,
+    sender uuid NOT NULL,
```
nick-w-nick, line 70:
```suggestion
    sender_id uuid NOT NULL REFERENCES users (id),
```

added user foreign key constraint + changed column name to `sender_id` for less ambiguity

scripts/init.sql
```diff
@@ -1,158 +1,147 @@
...
+    wishcard_id uuid NOT NULL,
```
nick-w-nick, line 71:
```suggestion
    wishcard_id uuid NOT NULL REFERENCES wishcards (id),
```

added wishcard foreign key constraint

scripts/init.sql
```diff
@@ -1,158 +1,147 @@
...
+    created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    CONSTRAINT messages_pkey PRIMARY KEY(id)
+);
+
+CREATE TABLE orders(
+    id uuid NOT NULL DEFAULT gen_random_uuid(),
+    status integer NOT NULL DEFAULT 0,
+    delivery_date timestamp,
+    tracking_info varchar(500),
+    donor_id uuid NOT NULL,
```
nick-w-nick, line 81:
```suggestion
    donor_id uuid NOT NULL REFERENCES users (id),
```

added user foreign key constraint

scripts/init.sql
```diff
@@ -1,158 +1,147 @@
...
+    child_id uuid NOT NULL,
```
nick-w-nick, line 82:
```suggestion
    child_id uuid NOT NULL REFERENCES children (id),
```

added child foreign key constraint

scripts/init.sql
```diff
@@ -1,158 +1,147 @@
...
+    item_id uuid NOT NULL,
```
nick-w-nick, line 83:
```suggestion
    item_id uuid NOT NULL REFERENCES items (id),
```

added items foreign key constraint

scripts/init.sql
```diff
@@ -1,158 +1,147 @@
...
+    agency_id uuid NOT NULL,
```
nick-w-nick, line 84:
```suggestion
    agency_id uuid NOT NULL REFERENCES agencies (id),
```
added agency foreign key constraint

scripts/init.sql
```diff
@@ -1,158 +1,147 @@
...
+    deleted_at timestamp,
+    CONSTRAINT users_pkey PRIMARY KEY(id)
+);
+
+CREATE TABLE verification_tokens(
+    id uuid NOT NULL DEFAULT gen_random_uuid(),
+    token varchar(255) NOT NULL,
+    "type" integer NOT NULL,
+    expiration timestamp NOT NULL,
+    user_id uuid NOT NULL,
```
nick-w-nick, line 112:
```suggestion
    user_id uuid NOT NULL REFERENCES users (id),
```

added user foreign key constraint

scripts/init.sql
```diff
@@ -1,158 +1,147 @@
...
+    state varchar(100) NOT NULL,
+    country varchar(100) NOT NULL,
+    zip_code varchar(50) NOT NULL,
+    status integer NOT NULL DEFAULT 0,
+    created_by uuid NOT NULL,
+    agency_id uuid NOT NULL,
+    child_id uuid NOT NULL,
+    item_id uuid NOT NULL,
+    image_id uuid NOT NULL,
+    order_id uuid NOT NULL,
```
nick-w-nick, line 140:
```suggestion
    agency_id uuid NOT NULL REFERENCES agencies (id),
    child_id uuid NOT NULL REFERENCES children (id),
    item_id uuid NOT NULL REFERENCES items (id),
    image_id uuid NOT NULL REFERENCES images (id),
    order_id uuid NOT NULL REFERENCES orders (id),
```

added agency, child, item, image and order foreign key constraints

scripts/init.sql
```diff
@@ -1,158 +1,147 @@
...
+    created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    CONSTRAINT images_pkey PRIMARY KEY(id)
+);
+
+CREATE TABLE items(
+    id uuid NOT NULL DEFAULT gen_random_uuid(),
+    "name" varchar(255) NOT NULL,
+    price numeric(10, 2) NOT NULL,
+    link varchar(2000) NOT NULL,
+    retailer varchar(255) NOT NULL,
```
nick-w-nick, line 59: Should we create a `retailers` table and reference the ID of the retailer instead of hardcoding the retailer name here each time?

scripts/init.sql
```diff
@@ -1,158 +1,147 @@
...
+    country varchar(100) NOT NULL,
+    zip_code varchar(50) NOT NULL,
+    verified boolean NOT NULL DEFAULT false,
+    employer_identification_number varchar(255) NOT NULL,
+    account_manager_id uuid NOT NULL,
+    image_id uuid NOT NULL,
+    created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    updated_at timestamp,
+    CONSTRAINT agencies_pkey PRIMARY KEY(id),
+    UNIQUE(employer_identification_number)
```
nick-w-nick, line 22:
```suggestion
    UNIQUE(id),
    UNIQUE(employer_identification_number)
```

@Enubia UUID collisions are next to impossible, but how do you feel about adding a unique constraint to each of them so we can prevent a duplicate entry if it somehow ever were to happen?

Enubia:
@nick-w-nick the constraints are already added via alter statements further down in the file

nick-w-nick:
> @nick-w-nick the constraints are already added via alter statements further down in the file

Ah, didn't see those, my view in VS Code was not displaying the rest of the file properly. Disregard the change suggestions then.

Can we change the `ON DELETE Cascade` rule to `ON DELETE RESTRICT` as we shouldn't be supporting hard deletes?

scripts/init.sql
stacysealky: We will use a default number EIN for the agencies created before the DB migration.
"12-3456789" add this to all existing verified agencies for now (format has to be exactly xx-xxxxxxx 9 numerical digits).

scripts/init.sql
stacysealky: I'm with @Enubia on this one. I'm in favor of just calling it id to avoid redundancy. 
We can use an Alias for join to avoid the ambiguous column name error, right? @nick-w-nick

scripts/init.sql
```diff
@@ -1,158 +1,147 @@
...
+CREATE TABLE users(
+    id uuid NOT NULL DEFAULT gen_random_uuid(),
+    first_name varchar(255) NOT NULL,
+    last_name varchar(255) NOT NULL,
+    email varchar(255) NOT NULL,
+    "password" varchar(100) NOT NULL,
+    "role" integer NOT NULL,
+    login_mode integer NOT NULL,
+    bio varchar(500),
+    verified boolean NOT NULL DEFAULT false,
```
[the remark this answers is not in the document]
stacysealky: in favor of this change also

scripts/init.sql
stacysealky: in favor of this change also

types/database.ts
stacysealky: agreed, ok let's keep it as meta_data

stacysealky (changes requested):
replied to comments above

scripts/init.sql
nick-w-nick: @stacysealky We should look into obtaining the EIN for all existing agencies for compliance reasons, honestly I think the best option is to remove the NOT NULL constraint until we can backfill the existing ones. I don't believe the `UNIQUE` constraint applies to `null` fields, so we should be fine there as long as that is the case.

scripts/init.sql
nick-w-nick: @stacysealky That is correct, this was just an attempt to prevent needing to do that, as it is against best practice to name them generically instead of by table, but I don't mind doing it this way instead.

scripts/init.sql
stacysealky: @nick-w-nick 
re: "We should look into obtaining the EIN for all existing agencies for compliance reasons"
-> I can look into it, but it will be kinda time consuming.

@Enubia we should remove the NOT NULL constraint for now

types/generated/database.ts
```diff
@@ -0,0 +1,166 @@
...
+  city: string;
+  state: string;
+  country: string;
+  zip_code: string;
+  verified: Generated<boolean>;
+  employer_identification_number: string | null;
+  account_manager_id: string;
+  image_id: string;
+  created_at: Generated<Timestamp>;
+  updated_at: Timestamp | null;
```
nick-w-nick, line 40:
```suggestion
  updated_at: Generated<Timestamp> | null;
```

I'm still trying to understand the point behind the `Generated` utility type, but wouldn't the `update_at` Timestamp also be considered `Generated`?

scripts/init.sql
```diff
@@ -1,187 +1,176 @@
...
+    image_id uuid NOT NULL,
+    created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    CONSTRAINT community_posts_pkey PRIMARY KEY(id)
+);
+
+CREATE TABLE images(
+    id uuid NOT NULL DEFAULT gen_random_uuid(),
+    url varchar(255) NOT NULL,
+    meta_data json,
+    created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
nick-w-nick, line 50:
```suggestion
    created_by uuid NOT NULL,
    created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
```

@Enubia Thoughts on adding a `created_by` field on the `images` table? We do technically need audit logging for a majority of our other tables anyways, but since images take up more resources than most things, it may be helpful if we can determine which user uploaded any given image in case we run into a situation where a bad actor tries to DOS us by spam uploading images. That way, we can apply a fix to prevent the attack and can also selectively delete all images uploaded by the individual.

We technically have a reference to the current image for the user/wishcard on those tables, but if they upload multiple images it breaks the 1 to 1 connection between them.

types/generated/database.ts
Enubia: I would not touch it since it is auto-generated and messing with it could potentially brick our typing system for the database

My guess as to why it's not auto-generated is that the type generation CLI does not take triggers into account

And technically you can still update that field manually

I generated that file into it's own output to see what we have, usually you would not do that

scripts/init.sql
Enubia: If we need it in more places wouldn't a revisions table make more sense?
This way we can set up a trigger I assume to log every interaction that has taken place by a user

Would also be useful for the back office so that we know who did what at any given time

stacysealky (approved):
Thank you @Enubia

types/generated/database.ts
nick-w-nick: @Enubia Didn't realize the file was generated initially, it not taking triggers into account is most definitely the issue there.

We will see how the types look once we are using them and can make changes if necessary later on.

scripts/init.sql
nick-w-nick: Yes, an audit table is usually very beneficial to have for a multitude of things, especially trailing during debugging as well as for notifications.

I don't think we should implement one yet, but it definitely would be an easy win post-kysely integration and database migration.

stacysealky: I think @nick-w-nick has a good point!

Enubia: I've added the created_by field with a one-to-many relation since a user can add images for an agency, wish card, child and himself

Enubia:
@nick-w-nick please review the last commit

scripts/init.sql
```diff
@@ -1,289 +1,228 @@
...
+    "updated_at" timestamp
+);
+
+CREATE TABLE "children" (
+    "id" uuid PRIMARY KEY NOT NULL DEFAULT (gen_random_uuid()),
+    "first_name" varchar(255) NOT NULL,
+    "last_name" varchar(255) NOT NULL,
+    "birth_year" integer NOT NULL,
+    "interest" varchar(255) NOT NULL,
+    "story" varchar(255) NOT NULL,
```
stacysealky, line 27: Could we change the story varchar to 2000 to cover edge cases with long stories?

scripts/init.sql
```diff
@@ -1,289 +1,228 @@
...
+    "agency_id" uuid NOT NULL,
+    "child_id" uuid NOT NULL,
+    "item_id" uuid NOT NULL,
+    "image_id" uuid NOT NULL,
+    "order_id" uuid NOT NULL,
+    "created_at" timestamp NOT NULL DEFAULT (CURRENT_TIMESTAMP),
+    "updated_at" timestamp
+);
+
+ALTER TABLE "verifications" ADD CONSTRAINT "verifications_user_id_fkey" FOREIGN KEY ("user_id") REFERENCES "users" ("id") ON DELETE RESTRICT;
```
nick-w-nick, line 133: @Enubia Have you confirmed that Kysely is able to pick up on these foreign key constraints since they are done outside of the initial table schema?

scripts/init.sql
```diff
@@ -1,289 +1,228 @@
...
+    "birth_year" integer NOT NULL,
+    "interest" varchar(255) NOT NULL,
+    "story" varchar(255) NOT NULL,
+    "image_id" uuid NOT NULL,
+    "created_at" timestamp NOT NULL DEFAULT (CURRENT_TIMESTAMP)
+);
+
+CREATE TABLE "community_posts" (
+    "id" uuid PRIMARY KEY NOT NULL DEFAULT (gen_random_uuid()),
+    "message" varchar(255) NOT NULL,
```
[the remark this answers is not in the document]
nick-w-nick: @stacysealky `VARCHAR(255)` doesn't mean it stores 255 characters, it means it's the max number of characters that can be counted by an 8-bit number, plus an additional byte as an offset.

[This StackOverflow post](https://stackoverflow.com/questions/1217466/is-there-a-good-reason-i-see-varchar255-used-so-often-as-opposed-to-another-l) describes it in a bit more detail, but tl;dr, if you don't know how much data you're going to be storing in a column but don't want to set a hard restriction, you use `VARCHAR(255)`.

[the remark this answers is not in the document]
Tokikko: Are you sure? Wont the string get truncated if it has over 255 characters? 

I would suggest change it to a text type or not specifying any length at all (so just varchar :) )

scripts/init.sql
Enubia: The types are generated by introspection of the created database where the constraints are already on the table themselves 

This means that kysely is aware of them

scripts/init.sql
Tokikko: In those cases maybe its better to leave it as text type without a length limit?

scripts/init.sql
nick-w-nick: Sounds good to me, just wanted to make sure that was the case.

Thanks!

scripts/init.sql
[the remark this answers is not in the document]
nick-w-nick: @Tokikko Actually, I think you're right, I may have gotten some wires crossed comparing it to SQL Server.

@Enubia I just checked some of my notes and recall speaking about this to a DBA and determining that `text` is a better type for any column storing a string of an undetermined length, as mentioned by @Tokikko.

Would you be open to swapping everything to `text` instead?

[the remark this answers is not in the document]
Enubia: Sure thing

nick-w-nick (approved):

nick-w-nick (changes requested):
Will approve once all varchars are updated to `text`

scripts/init.sql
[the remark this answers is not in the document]
stacysealky: wait wait, but we do want to have a max limit for all string inputs in db, right?
I mean there's a frontend validation for each string input. For example, in the form, it allows 255 chars max for first name, 2000 for the child story, etc. 
So shouldn't we want to keep the rules consistent?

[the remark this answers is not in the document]
nick-w-nick: @stacysealky the parameter used for `VARCHAR` is for bytes, not characters, so we would need to add some arbitrary buffer on top of whatever we set the limit to. It would also likely be quite bothersome to change it later on if we changed the validation rules.

We will implement length validation rules on the backend to prevent entries that exceed the max limit as a secondary safeguard.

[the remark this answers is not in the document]
stacysealky: Ok sounds good. I'm in favor of changing all to TEXT.

Enubia:
@nick-w-nick @stacysealky updated varchar to text in all places that made sense

scripts/init.sql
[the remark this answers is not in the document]
Enubia: @nick-w-nick 

varchar and text do not really differ in its type

> The PostgreSQL Text data type is used to keep the character of infinite length. And it can hold a string with a maximum length of 65,535 bytes.
> It is the same as the PostgreSQL Varchar data type, rather without any argument for mentioning the length of the character ie., the PostgreSQL Varchar Data Type without n is the same as the Text data type.

PR #622, opened 2023-08-13
fixed the profile page to match prod

### LINKED ISSUE #613: DEFECT: Profile page needs to match prod
screencapture-donate-gifts-profile-2023-08-10-18_52_05

Please make sure profile page matchs the prod

stacysealky:
resolves #613


js/react/components/nav/index.jsx
```diff
@@ -47,45 +43,32 @@ function Nav(props) {
 								Community
 							</a>
 						</li>
-						{/* No user */}
-						{!user && (
+						{!user ? (
```
Enubia, line 46: Please changes this back to what it was before

Nested ternary conditions are not readable and overall a bad convention

@NMilivojevic did it the same way as you do now and I requested from her to change it as well

Enubia (changes requested):

js/react/components/nav/index.jsx
stacysealky: Ok I will separate them, but the text display before was not what we wanted, so I changed it.

js/react/components/nav/index.jsx
```diff
@@ -47,45 +43,33 @@ function Nav(props) {
...
-									</a>
-								</li>
-								<li className="nav-item">
-									<a className="nav-link" href="/wishcards/create">
-										Create a Wish Card
-									</a>
-								</li>
-							</>
-						)}
+						) : null}
```
Enubia, line 72: you can actually remove the null defaults here and change it to a condition chain
```suggestion
						{user?.userRole === 'partner' && agency?.isVerified && (
							<li className="nav-item">
								<a className="nav-link" href="/wishcards/create">
									Create a Wish Card
								</a>
							</li>
						)}
						{user?.userRole === 'admin' && (
							<li className="nav-item">
								<a className="nav-link" href="/admin/">
									Admin Panel
								</a>
							</li>
						)}
```

stacysealky: @Enubia  - Oh I try to avoid && chains whenever possible because I was told it's not a good practice by a senior react dev at work.
It's to prevent '0' from showing up. For example, if you do {someData.length && (<></>)} this can show a 0 when the condition does not meet. I was told it's not a good habit, as it can cause unwanted defect in some cases.

NMilivojevic: @stacysealky thanks for sharing that, good to know, https://stackoverflow.com/questions/53048037/react-showing-0-instead-of-nothing-with-short-circuit-conditional-component, will look more into it

Enubia: > @Enubia - Oh I try to avoid && chains whenever possible because I was told it's not a good practice by a senior react dev at work. It's to prevent '0' from showing up. For example, if you do {someData.length && (<></>)} this can show a 0 when the condition does not meet. I was told it's not a good habit, as it can cause unwanted defect in some cases.

Ah ok, wasn't aware of that. Well, it's ugly but I can see the benefit. Please ignore my suggestion in that case.

PR #627, opened 2023-08-17
Modal content for login, login with email and signup

NMilivojevic:
Header.jsx was created and mounted in master.pug, it contains Nav and Auth components and modal content state (login, login with email, signup) in reducer with Auth component being the enter point for our different modals like login, login with email, signup, etc. One modal content is missing which is signup with email since its a flow on its own and this PR is already big enough :D. The bootstrap modal open/close state is controlled using the reducer in Header.jsx, it is initially opened in Nav.jsx, and in Auth.jsx depending on which state is true showLogin, showLoginWithEmail, showSignup respective components are rendered which are separated into login and signup folders in the auth folder. Google and facebook  buttons are TBD. Css is added in _modal.scss, and these rounded login, signup buttons are added in _buttons.scss. Login click in the navbar opens the modal, but /login route is still rendering the old login page until we have the modal fully functional. Also let me know if there is anything in my git workflow which i should change, these old commits showing on every PR for example, if I can change that on my end, please let me know or anything else.


Enubia (approved):
LGTM!

PR #662, opened 2023-08-24
Fix: Messaging not working

Jacobjeevan:
- Frontend was passing in the user and wishcard Ids while backend was expecting the object.
   - Alternatively, could change the backend code (add in extra db calls for fetching user/wishcard in postMessageValidationRules and handlePostMessage) and just pass in the Ids from frontend.


stacysealky (approved):
Looks good

Enubia (approved):

PR #667, opened 2023-09-06
Added check for agency verification when rendering the submission form

Tokikko:


js/react/components/community/index.jsx
```diff
@@ -6,7 +6,7 @@ import Masonry, { ResponsiveMasonry } from 'react-responsive-masonry';
 import LoadingCard from '../shared/LoadingCard.jsx';
 
 function Community(props) {
-	const { user, _csrf } = props;
+	const { user, agency, _csrf } = props;
```
stacysealky, line 9: what is the _csrf for?

src/controller/community.ts
```diff
@@ -20,10 +20,17 @@ export default class CommunityController extends BaseController {
...
+			const { user } = res.locals;
 			const posts = (await this.postRepository.getAllPostsByVerifiedAgencies()).sort(
 				(a, b) => a.createdAt.getTime() - b.createdAt.getTime(),
 			);
-			this.renderView(res, 'pages/community', { posts });
+
+			if (user?.userRole !== 'partner') {
+				return this.renderView(res, 'pages/community', { posts });
+			}
+
```
stacysealky, line 31: Can you add a condition so that if agency.isVerified it returns {posts, agency} and else {posts} even for user.userRole == 'partner'

stacysealky (changes requested):
Looks good. Just a minor suggestion

js/react/components/community/index.jsx
Tokikko: ITs a token to prevent cross site requests. Seems its set for a future use so i left it.

src/controller/community.ts
Tokikko: I can but its better this way. If its not a partner agency is not necessary and we have to check if its verified on the frontend anyway.

stacysealky (approved):

PR #673, opened 2023-09-12
fix: email verification page

### LINKED ISSUE #624: DEFECT: email verification broken
Issue 1. Email verification link is broken and shows the 500 error page
Screenshot 2023-08-14 at 2 55 59 PM

Issue 2. Automated email has broken logo image
Screenshot 2023-08-14 at 2 56 27 PM

Issue 3. Need to add "Resend verification link" option for email verification, in case the first one didn't send or not working.
Please add another button under "Settings" card in profile page. 
We will redesign the settings section later, so for now just add the same button style.
Also, this button is only visible if user.isVerified === false
Screenshot 2023-08-14 at 3 01 34 PM

Enubia:
resolves #624

### DIFF
--- a/src/controller/profile.ts
+++ b/src/controller/profile.ts
@@ -236,7 +236,6 @@ export default class ProfileController extends BaseController {
 
 				return this.renderView(res, 'profile/overview', {
 					user,
-					agency,
 					wishCards,
 					wishCardsLength,
 					draftWishcards,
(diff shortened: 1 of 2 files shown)


Enubia:
@stacysealky yes, so, what happened was that the user was verified, but we attached an empty agency object even tho the user role was set to donor which caused the template to crash because we are checking for a property in the modal section

i removed the agency object in case the role is donor during verification
tested it and it's working again as intended

Issue #616, opened 2023-08-12
STORY: create a migration script from mongoDB -> postgres (remap seeder jsons)


stacysealky:
-Dependent on completion of #615 

- Use uuid for primary key
https://stackoverflow.com/questions/33274291/uuid-or-sequence-for-primary-key

- Change all constraints from integer to varchar(255)~

- Check the diagram to convert the seeder json data to match the new table definitions

https://dbdiagram.io/d/64ddb26d02bd1c4a5eea4e48

![image](https://github.com/donategifts/donategifts/assets/44695100/3c17b264-03df-4cdb-9ed3-eef9e3301d1b)

Enubia:
> Use uuid for primary key
> Change all constraints from integer to varchar(255)

done in #615

nick-w-nick:
> Check the diagram to convert the seeder json data to match the new table definitions

done in #678

PR #678, opened 2023-09-13
Added Postgres Seeder + Schema Updates

nick-w-nick:
In this PR, I've added a brand new seeder to generate and insert data into Postgres based on our SQL schema.

This seeder comes with modular functionality, splitting the data generation, processing, and seeding into separate steps allowing for easy additions and modifications as necessary. The seeder even detects and updates stale references to records that no longer exist/match the relation criteria.

I've also made numerous adjustments to the SQL schema, most of which we've already discussed previously. Please let me know if you have any questions regarding the changes and/or have any suggestions.

(this PR addresses #616)


scripts/init.sql
```diff
@@ -8,14 +14,14 @@ CREATE TABLE "agencies" (
     "address_line_2" varchar(255),
     "city" varchar(100) NOT NULL,
     "state" varchar(100) NOT NULL,
-    "country" varchar(100) NOT NULL,
+    "country_code" varchar(100) NOT NULL,
```
nick-w-nick, line 17: Changed to `country_code` for simplified standardization vs. storing the full country string.

scripts/init.sql
```diff
@@ -8,14 +14,14 @@ CREATE TABLE "agencies" (
...
     "zip_code" varchar(50) NOT NULL,
-    "verified" boolean NOT NULL DEFAULT false,
+    "is_verified" boolean NOT NULL DEFAULT false,
```
nick-w-nick, line 19: changed to `is_verified` to make it less ambiguous that it's a boolean

scripts/init.sql
```diff
@@ -8,14 +14,14 @@ CREATE TABLE "agencies" (
...
     "employer_identification_number" varchar(100) UNIQUE,
     "account_manager_id" uuid NOT NULL,
-    "image_id" uuid NOT NULL,
-    "created_at" timestamp NOT NULL DEFAULT (CURRENT_TIMESTAMP),
-    "updated_at" timestamp
+    "image_id" uuid,
+    "created_at" timestamptz NOT NULL DEFAULT (CURRENT_TIMESTAMP),
```
nick-w-nick, line 23: changed all timestamps to `timestamptz` as I found great difficulty in getting Postgres to use the same timezone for everything regardless of system time, so it would be wise to store the timezone along with the timestamp for easier conversions in-code.

scripts/init.sql
```diff
@@ -25,90 +31,81 @@ CREATE TABLE "children" (
     "birth_year" integer NOT NULL,
     "interest" text NOT NULL,
     "story" text NOT NULL,
-    "image_id" uuid NOT NULL,
-    "created_at" timestamp NOT NULL DEFAULT (CURRENT_TIMESTAMP)
+    "image_id" uuid,
+    "agency_id" uuid NOT NULL,
```
nick-w-nick, line 35: added `agency_id` to child to link them to an agency

scripts/init.sql
```diff
@@ -25,90 +31,81 @@ CREATE TABLE "children" (
...
+    "status" OrderStatus NOT NULL DEFAULT ('pending'::OrderStatus),
+    "delivery_date" timestamptz,
     "tracking_info" varchar(500),
     "donor_id" uuid NOT NULL,
-    "child_id" uuid NOT NULL,
-    "item_id" uuid NOT NULL,
-    "agency_id" uuid NOT NULL,
-    "created_at" timestamp NOT NULL DEFAULT (CURRENT_TIMESTAMP),
-    "updated_at" timestamp
+    "wishcard_id" uuid NOT NULL,
```
nick-w-nick, line 81: removed all unnecessary relations to the child/agency/item when they can instead be retrieved by checking the associated tables via the `wishcard_id`.

scripts/mongoimport.sh
Enubia, line 1: please keep this file for now until we migrated fully to postgres on live
we still need it to pull the live data from mongo so that we can migrate the db entries

Enubia (changes requested):
Really good implementation, I like the workflow approach 👍 
Just one small thing, I'd like to keep the mongoimport script since we will need it for the migration

scripts/mongoimport.sh
nick-w-nick: @Enubia The target branch has been updated and the mongo seeder has been re-added, please see the [`scripts/mongo-seeder`](https://github.com/donategifts/donategifts/tree/1649cb97c79dd633c382fd55ca8b0155688ad3ef/scripts/mongo-seeder) directory.

I've also added a script to the package.json called `seed:mongo` that automatically runs the seeder so you don't have to navigate to the seeder directly.

Enubia (approved):

Issue #558, opened 2023-07-26
STORY: convert home sample cards to react component


Enubia:
convert it from a pug template to a react component

you can take the rendering from [/js/react/components/wishcards](https://github.com/donategifts/donategifts/blob/master/js/react/components/wishcards/index.jsx)

what would be good is to create a shared wish card component and use that in both react components

PR #682, opened 2023-09-26
Convert home sample cards to react component

NMilivojevic:
I have created `WishCardsCarousel.jsx` component in `home` folder. Bootstrap classes from `homeSampleCards.pug` are used there. 

A separate `WishCard.jsx` component is made in `shared` folder and it returns only the card with image and text below it. 
This component is used in `wishcards/index.jsx` and in `home/WishCardsCarousel.jsx`. It receives as props the full `wishCard` object.

In place of `homeSampleModal.pug` in `home.pug` page, is a `#carousel-root,` which is used below in js to mount `WishCardsCarousel.jsx`. The wishCards response from backend is passed to it. Should the user object response be passed to it as well?


Enubia:
> Should the user object response be passed to it as well?

I don't think it's needed

Enubia (approved):
LGTM! @stacysealky please check it as well

NMilivojevic:
> > Should the user object response be passed to it as well?
> 
> I don't think it's needed

There was one instance of conditional rendering based if logged in user clicked on `Donate Gift` button, or public user did. In `homeSampleCards.pug` it went like this
```
if user
a( href='wishcards/donate/' + currentCard._id )
button.....Donate Gift
else
button.... Donate Gift
                                                        
```

Not sure if that conditional rendering is still relevant.

js/react/utils/helpers.jsx
```diff
@@ -6,3 +6,11 @@ export const loadlGoogleClientLibraryScript = (src) => {
...
 	body.appendChild(tag);
 };
+
+export const chunkArray = (arr, chunkSize) => {
+	const chunkedArray = [];
+	for (let i = 0; i < arr.length; i += chunkSize) {
+		chunkedArray.push(arr.slice(i, i + chunkSize));
+	}
+	return chunkedArray;
+};
```
nick-w-nick, line 16:
```suggestion
export const chunkArray = (initialArray, chunkSize) => {
	const chunks = [];
	for (let index = 0; index < initialArray.length; index += chunkSize) {
		const chunk = initialArray.slice(index, index + chunkSize);
		chunks.push(chunk);
	}
	
	return chunks;
};
```

Minor readability improvements

Enubia:
> > > Should the user object response be passed to it as well?
> 
> > 
> 
> > I don't think it's needed
> 
> 
> 
> There was one instance of conditional rendering based if logged in user clicked on `Donate Gift` button, or public user did. In `homeSampleCards.pug` it went like this
> 
> ```
> 
> if user
> 
> a( href='wishcards/donate/' + currentCard._id )
> 
> button.....Donate Gift
> 
> else
> 
> button.... Donate Gift
> 
>                                                         
> 
> ```
> 
> 
> 
> Not sure if that is still relevant.

I think for now it's not needed, if we find out during testing that it should be added again we can do that in a new task

stacysealky:
@NMilivojevic Looks great!! Will merge after the above comments are addressed. Thanks!

Enubia:
resolves #558

PR #703, opened 2023-10-05
fix: carousel clipping and wishcard style

### LINKED ISSUE #699: DEFECT: fix clipping in wishcard carousel
image

additionally, check what's happening when the next section is loaded because between them the container seems to stretch to the full view width which makes it look ugly for a brief second

Enubia:
resolves #699


stacysealky (approved):
Nice work!

PR #705, opened 2023-10-06
refactor: controllers

Enubia:


Enubia:
@nick-w-nick @stacysealky 
some of the repository functions will find no usage or will be updated to a more generic approach, for now, we only ported what we thought might make sense.

When we update the views we will also need to update a bunch of the API calls and therefore also update the queries.

PR #725, opened 2023-10-14
Messaging Fixes

Jacobjeevan:
- Disabled custom messaging option for now
	- Since we are progressing with Db migration and stories, better to add custom messaging afterwards,
rather than modifying existing mongo objects to allow for custom messaging and doing the same work later.
- Added verified user check to display messaging option

Note: Prob do squash and merge since I merged master into this branch.


Enubia (approved):
LGTM!

PR #737, opened 2023-10-26
feat: added basic mantine theme

Enubia:


js/react/utils/mantineProviderWrapper.jsx
```diff
@@ -1,7 +1,81 @@
 import { MantineProvider, createTheme } from '@mantine/core';
 
 const theme = createTheme({
-	/** Put your mantine theme override here */
```
stacysealky, line 4: Are these the colors that we are already using?

js/react/pages/admin/routes/agency/overview.jsx
```diff
@@ -40,6 +50,12 @@ export default function Overview() {
 		}
 	}, [value]);
 
+	useEffect(() => {
```
stacysealky, line 53: Is this how you want to handle frontend navigation from now on? 
I previous used window for this in WishCardCreate.jsx. 
setTimeout(() => window.location.replace('/wishcards/manage'), 1000);

Should we use both or unify the redirect method for code consistency?

If we show a success toast message, we should probably let it display for a second before redirecting the page, right?

stacysealky (approved):
Approved, but please check my comments :)

js/react/utils/mantineProviderWrapper.jsx
Enubia: I actually have to check them again, I noticed that the mantine generator does not properly use the input colors, so I'll update them again

js/react/pages/admin/routes/agency/overview.jsx
Enubia: We can use either, this is just the recommended way from the react router docs

Enubia:
@stacysealky updated the colors, the darkest version of each of them is the one we use in the design

PR #742, opened 2023-11-16
Curated carousel

stacysealky:
Curated carousel displays when there is 0 wishcard to be donated, so that the home page looks better.

### DIFF
--- a/src/db/repository/WishCardRepository.ts
+++ b/src/db/repository/WishCardRepository.ts
@@ -9,7 +9,6 @@ import WishCard from '../models/WishCard';
 
 import DonationRepository from './DonationRepository';
 import UserRepository from './UserRepository';
-
 export default class WishCardRepository {
 	private wishCardModel: typeof WishCard;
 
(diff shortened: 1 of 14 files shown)


stacysealky:
It's throwing error because of the `curated.push(card);` @Enubia

Enubia:
> It's throwing error because of the `curated.push(card);` @Enubia 

Yes, because your are missing a proper type

src/db/repository/WishCardRepository.ts
```diff
@@ -50,6 +49,32 @@ export default class WishCardRepository {
...
+			'5f7f8a8eb09ed5121216bd1b',
+			'5fc48548495483e2e4e87432',
+			'5f90a7fc987faa6593d8c061',
+			'5fd18775e7553d5923e6ce63',
+			'5fd188fee7553d5923e6ce64',
+			'5f9b3373987faa6593d8c07a',
+			'5fc486a5495483e2e4e87433',
+			'5f847114a99508e349bb87fd',
+		];
+		const curated = [];
```
Enubia, line 68:
```suggestion
		const curated: WishCard[] = [];
```

Enubia: this should fix the build error

Enubia (changes requested):

Enubia (approved):

Issue #747, opened 2023-11-18
Fix the broken external links for agency partners (only some broken)


stacysealky:
On wishcards/single page, the agency website is broken in some cases. Please fix them.
<img width="1539" alt="Screenshot 2023-11-17 at 6 18 16 PM" src="https://github.com/donategifts/donategifts/assets/26069780/3acee242-aeaf-4646-9e49-fc9f130e10d3">

DeanRTaylor1:
Hey, I'm working on this now.

PR #763, opened 2023-11-22
Add validation for url protocol

DeanRTaylor1:
The issue was caused by not having a protocol component in the URL and thus the browser believing it to be a relative href tag. 

I added a static utility function as this seems to be the preferred approach in your codebase. There are a couple of other ways to do it so if you are unhappy with the implementation or modifying the agency object directly, let me know.

Lastly, I only updated the wishcard controller as it was the only page mentioned. A more robust fix would be to ensure that http:// is added when creating the agency but that would have wider implications so I'll let you decide whether or not to do that.

### DIFF
--- a/src/controller/wishcard.ts
+++ b/src/controller/wishcard.ts
@@ -414,6 +414,9 @@ export default class WishCardController extends BaseController {
 
 			// this agency object is returning undefined and breaking frontend
 			const agency = wishcard!.belongsTo;
+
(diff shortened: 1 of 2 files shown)


DeanRTaylor1:
Relates to #747

Enubia (approved):
LGTM!

I agree that this should also be handled when the agency is created, but since we are also working on a complete rewrite of the backend we'll handle it later on.

PR #795, opened 2023-12-01
Added tracking info

Enubia:
Donations page on the admin view was missing a tracking info input field.
Field was added and some refactoring was done to make usability on mobile easier as well as switching the default admin view

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,1 @@
-unsafe-perm=true
+registry=https://registry.npmjs.org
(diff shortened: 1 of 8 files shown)


Jacobjeevan:
LGTM 👍 

Minor comment: Not a big issue, but possible to clear an entry? (i.e. update the tracking field in case I remove the existing one). Though it's not a big deal since I can add space characters and that works.

Jacobjeevan (approved):

Enubia:
> LGTM 👍
> 
> Minor comment: Not a big issue, but possible to clear an entry? (i.e. update the tracking field in case I remove the existing one). Though it's not a big deal since I can add space characters and that works.

There should probably never be an empty entry, only on the initial creation of the donation in the database.

PR #832, opened 2023-12-05
Donation history enhancements

stacysealky:
Fixed a few small issues

- donation amount on donation history page only shows max of 2 decimal points to prevent edge case defects
- added the tshirt image and modified the popover for donation karma points
- minor design change (more margin bottom in mobile screen)


js/react/bundle.jsx
```diff
@@ -35,7 +35,7 @@ window.AgencyRegister = AgencyRegister;
 import Contact from './pages/contact.jsx';
 window.Contact = Contact;
 
-import DonationHistory from './pages/profile/donationhistory.jsx';
+import DonationHistory from './pages/profile/donation.jsx';
```
Enubia, line 38: why did you rename it? was `donationhistory` not correct?

js/react/pages/payment/success.jsx
```diff
@@ -83,9 +85,12 @@ function PaymentSuccess() {
...
 							As a result of this donation, you have earned{' '}
-							<a href="/profile/donations" target="_blank">
-								{10 + Math.ceil(successInfo?.amount * 2)} donation karma points.
-							</a>
+							<b>{10 + Math.ceil(successInfo?.amount * 2)}</b> donation karma points.
+							<PopOver
+								text={Translations.DONATION_HISTORY.karmaPoints.popOverText}
+								isImgProvided={true}
+								imgSrc="/img/tshirt-2023.png"
+							/>
```
Enubia, line 93: please check the changes in master, I've reworked this component a bit since a few fields were redundant

js/react/pages/profile/donation.jsx
```diff
@@ -31,11 +31,17 @@ function DonationHistory() {
 	useEffect(() => {
 		if (donations) {
 			setTotalGifts(donations.length);
+
+			let newTotalDonation = Number(0.0);
```
Enubia, line 35: I think the `Number` constructor can be removed since you already assign `0.0` as a value

Enubia (changes requested):

js/react/bundle.jsx
stacysealky: @Enubia yeah but it was not camel case haha so I wanted to match the current naming conventions

Enubia: oh but pages are fine with all lowercase :D

component file names should be PascalCase

Enubia (approved):

PR #823, opened 2023-12-03
Search and Filtering

Jacobjeevan:
Added search bar and filtering options, may need additional styling though, but functionality is good to go.

image

PR #833, opened 2023-12-05
810 searchbar component

Jacobjeevan:
#823 
 +

- added logic to rerender when user clicks on any of the filtering options (rather than having to click submit).
- only display agencies with wishcards in the agency sort dropdown (and ordered them alphabetically)
- Added no results message for now (maybe do additional styling?)

image

Note: Still has to use submit button for the text input ofc.

### DIFF
--- a/js/react/pages/wishCards.jsx
+++ b/js/react/pages/wishCards.jsx
@@ -1,7 +1,7 @@
 import { TextInput, Radio, Group, Checkbox, Select, Switch } from '@mantine/core';
 import axios from 'axios';
 import PropTypes from 'prop-types';
-import { useEffect, useState } from 'react';
+import { useEffect, useState, useRef } from 'react';
 
 import LoadingCard from '../components/shared/LoadingCard.jsx';
 import WishCard from '../components/shared/WishCard.jsx';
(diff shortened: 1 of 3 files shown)


js/react/pages/wishCards.jsx
```diff
@@ -7,57 +9,181 @@ import MantineProviderWrapper from '../utils/mantineProviderWrapper.jsx';
...
+				agencyFilter: searchQueryParams.agencyFilter
+					? agencies.find((agency) => agency.agencyName == searchQueryParams.agencyFilter)
+							._id
+					: null,
+			},
+			{
+				headers: {
+					'Content-Type': 'application/json',
+				},
+			},
```
Enubia, line 84: headers can be omitted if it's not a formdata object or you need to include some special headers for the request

js/react/pages/wishCards.jsx
```diff
@@ -7,57 +9,181 @@ import MantineProviderWrapper from '../utils/mantineProviderWrapper.jsx';
...
 			<div id="wishcards" className="bg-light p-4">
-				<div className="container">
+				<div className="container d-flex flex-column">
+					<div className="d-flex justify-content-center">
+						<form className="p-4 rounded-2 bg-white" onSubmit={handleSearchSubmit}>
+							<div className="d-flex flex-wrap gap-3 mb-2">
+								<TextInput
+									placeholder="Search"
+									value={searchTextInput}
+									onChange={handleSearchInputChange}
```
Enubia, line 99: instead of an `onChange` event you could use a `ref` here so you don't fire an event everytime the input changes

Enubia (changes requested):

js/react/pages/wishCards.jsx
Jacobjeevan: Done

Enubia (approved):
LGTM!

PR #834, opened 2023-12-05
feat: added redirect after login to donation page

Enubia:


Jacobjeevan (approved):
LGTM 👍  I only tested with regular login (I didn't have the google client, but I trust it works).

PR #840, opened 2023-12-07
Converted Partner Carousel to Mantine

Jacobjeevan:
There's a small difference between this and prod ver (the navigate buttons on production quickly jumps to next slide, while Mantine's does the transition animation; we could change the speed variable, but that would affect the animation as well).

### DIFF
--- a/js/react/components/home/PartnerCarousel.jsx
+++ b/js/react/components/home/PartnerCarousel.jsx
@@ -5,7 +5,7 @@ import { useRef } from 'react';
 
 import { PARTNER_CAROUSEL_DATA } from '../../utils/constants';
 
-function PartnerCarousel() {
+export default function PartnerCarousel() {
 	const autoplay = useRef(Autoplay({ delay: 2500 }));
 
 	const partnerCarouselSlides = PARTNER_CAROUSEL_DATA.map((data) => (
@@ -14,7 +14,7 @@ function PartnerCarousel() {
 			className="d-flex flex-column justify-content-center align-items-center"
 		>
 			<a
-				href={data.url ? data.url : '#'}
+				href={data.url || '#'}
 				target={data.url ? '_blank' : undefined}
 				rel={data.url ? 'noreferrer' : undefined}
 			>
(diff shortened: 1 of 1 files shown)


js/react/components/home/PartnerCarousel.jsx
```diff
@@ -0,0 +1,51 @@
+import { Carousel } from '@mantine/carousel';
+import { Image, MantineProvider } from '@mantine/core';
+import Autoplay from 'embla-carousel-autoplay';
+import { useRef } from 'react';
+
+import { PARTNER_CAROUSEL_DATA } from '../../utils/constants';
+
+function PartnerCarousel() {
```
Enubia, line 8: you can do the export in this line, there are no proptypes applied on the component

js/react/components/home/PartnerCarousel.jsx
```diff
@@ -0,0 +1,51 @@
...
+	const autoplay = useRef(Autoplay({ delay: 2500 }));
+
+	const partnerCarouselSlides = PARTNER_CAROUSEL_DATA.map((data) => (
+		<Carousel.Slide
+			key={data.img}
+			className="d-flex flex-column justify-content-center align-items-center"
+		>
+			<a
+				href={data.url ? data.url : '#'}
```
Enubia, line 17:
```suggestion
				href={data.url || '#'}
```

js/react/components/home/PartnerCarousel.jsx
```diff
@@ -0,0 +1,51 @@
...
+				rel={data.url ? 'noreferrer' : undefined}
+			>
+				<Image src={data.img} fit="cover" w={150} h="auto" alt={data.alt} />
+			</a>
+		</Carousel.Slide>
+	));
+
+	return (
+		<MantineProvider>
+			<div className="container my-5">
```
Enubia, line 28: you can use the `Container` component from mantine here although I'm not sure why we need a container if the carousel should span across the whole screen

Enubia (changes requested):

Enubia (approved):
LGTM!

PR #850, opened 2023-12-08
Admin: Wishcard Images Edit

Jacobjeevan:
Edit both child and wishItem images on administration panel. The wishcard state has all the other props, in case we need to make all the other fields editable on admin end.

image

### DIFF
--- a/js/react/pages/admin/routes/wishcard/administration.jsx
+++ b/js/react/pages/admin/routes/wishcard/administration.jsx
@@ -13,12 +13,9 @@ export default function Administration() {
 		type: '',
 	});
 	const [isToastVisible, setIsToastVisible] = useState(false);
-	const [shouldNavigate, setShouldNavigate] = useState(false);
-	const [wishcardId, setWishcardId] = useState('');
 
 	const forward = (id) => {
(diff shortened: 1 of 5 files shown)


Jacobjeevan:
Note: converting to draft as I need to ensure toast works (right now, it just displays "this is toast" on save).

js/react/pages/admin/routes/wishcard/administration.jsx
```diff
@@ -57,6 +66,12 @@ export default function Administration() {
 		});
 	}, []);
 
+	useEffect(() => {
+		if (shouldNavigate) {
+			navigate(`/wishcards/${wishcardId}`);
+		}
```
Enubia, line 72: I think this can be moved to the forward function, there's no need to listen for a change

Enubia (changes requested):

js/react/pages/admin/routes/wishcard/administration.jsx
Jacobjeevan: Good point 👍

Enubia (approved):

Issue #752, opened 2023-11-20
Convert profile (My Account) page to React


stacysealky:
Current profile.pug (My Account page) needs to be converted into React.

PR #854, opened 2023-12-19
React migration: My Account

Jacobjeevan:
#752 

The old overview page content was moved to oldOverview.pug, so if you want to compare quickly, modify handleGetIndex in [src/controller/profile.ts](../blob/89f16c4c5313f3bc77923d1c8162e43d312295d7/src/controller/profile.ts) to render 'profile/oldOverview' instead of 'profile/overview'.

We can remove the oldOverview page once PR is ready to be merged.


js/react/components/profile/accountBioEditModal.jsx
Enubia, line 1: Components should be written in `PascalCase`

package.json
```diff
@@ -43,6 +43,7 @@
 	},
 	"homepage": "https://github.com/stacysealky/donategifts#readme",
 	"dependencies": {
+		"@mantine/form": "7.3.1",
```
Enubia, line 46: please move this to dev deps

Enubia (changes requested):
Will check it out locally this week

